PR #6, opened 2019-11-12
fix: Make lerna publish work

amaurymartiny:
@yjkimjunior Could you merge and try `yarn lerna:publish`, I don't think lerna:version is necessary here (but i might be wrong)


yjkimjunior:
it says no tagged release found, so I guess it needs a small diff and lerna:version bump

PR #7, opened 2019-11-12
chore: Clean up package.jsons

amaurymartiny:
- same author, license etc in package.jsons
- outputs js files in `lib/` instead of `build/`. In this case, users would do:
```diff
import { NavButton } from '@substrate/ui-components';
- import { NavButton } from '@substrate/ui-components/build/NavButton';
+ import { NavButton } from '@substrate/ui-components/lib/NavButton';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,25 +1,24 @@
 {
   "name": "nomidot",
-  "version": "1.0.0",
-  "main": "index.js",
-  "author": "",
+  "author": "Parity Technologies <admin@parity.io>",
+  "description": "Staking Portal for Polkadot",
   "license": "Apache-2.0",
+  "private": true,
+  "repository":"https://github.com/yjkimjunior/Nomidot",
+  "scripts": {
+    "prelerna:publish": "lerna exec yarn build --stream --scope @substrate/context --scope @substrate/ui-components",
+    "lerna:publish": "lerna version --conventional-commits && lerna publish from-git",
+    "lint": "tsc --noEmit",
+    "start": "cd ./front/gatsby && gatsby develop",
+    "storybook": "cd ./front/ui-components && yarn storybook"
+  },
   "workspaces": {
     "packages": [
       "front/*"
     ]
   },
-  "private": true,
   "devDependencies": {
     "lerna": "^3.18.4",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,25 +1,24 @@
...
-  "main": "index.js",
-  "author": "",
+  "author": "Parity Technologies <admin@parity.io>",
+  "description": "Staking Portal for Polkadot",
   "license": "Apache-2.0",
+  "private": true,
+  "repository":"https://github.com/yjkimjunior/Nomidot",
+  "scripts": {
+    "prelerna:publish": "lerna exec yarn build --stream --scope @substrate/context --scope @substrate/ui-components",
+    "lerna:publish": "lerna version --conventional-commits && lerna publish from-git",
```
amaurymartiny: So after this PR, we would just run: `yarn lerna:publish`. It will:
- build all the packages
- do lerna version && lerna publish

it'll still be interactive, so we can cancel if we see something's wrong.

package.json
```diff
@@ -1,25 +1,24 @@
...
-  "version": "1.0.0",
-  "main": "index.js",
-  "author": "",
+  "author": "Parity Technologies <admin@parity.io>",
+  "description": "Staking Portal for Polkadot",
   "license": "Apache-2.0",
+  "private": true,
+  "repository":"https://github.com/yjkimjunior/Nomidot",
+  "scripts": {
+    "prelerna:publish": "lerna exec yarn build --stream --scope @substrate/context --scope @substrate/ui-components",
```
yjkimjunior: small thing but could we just make this `lerna:pre` and `lerna:publish`?

amaurymartiny: hmm, it's like a package.json convention, if you a script `{s}` and another `pre{s}`, then `yarn {s}` will run `yarn pre{s}` first

amaurymartiny: It's basically to save us from running 2 scripts to publish

amaurymartiny: I could also put everything into one long script, if that's clearer to you

yjkimjunior: ah okay, i didn't know that, let's jsut leave it as is in that case

PR #12, opened 2019-11-15
Onboard with Extension

yjkimjunior:
browser should not hold keyring, only use injected accounts


front/gatsby/src/pages/onboarding.tsx
```diff
@@ -2,16 +2,27 @@
...
 
-export const Onboarding = () => {
+const ONBOARDING_STEPS = [
+  'stash',
+  'controller'
+]
+
+interface Props extends RouteComponentProps {};
+
+export const Onboarding = (props: Props) => {
```
amaurymartiny: Could you add function return types?

We can add a stricter eslint later, but let's give our future selves less work

front/context/src/AccountsContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+interface Props {
+  children: any;
+}
+
+export function AccountsContextProvider (props: Props): React.ReactElement {
+  const { children } = props;
+  const [injectedAccounts, setInjected] = useState<InjectedAccountExt[]>([] as InjectedAccountExt[]);
+
+  const getInjected = async () => {
```
amaurymartiny: Could put `getInjected` in the useEffect, so that it only gets created once

front/gatsby/src/pages/index.tsx
```diff
@@ -2,23 +2,29 @@
...
   return (
     <ContextGate>
       <ApiContext.Consumer>
         {({ isReady }) => (
           isReady && (
             <Layout>
               <SEO title="Home" />
-              <Onboarding />
+              <Router>
+                <Onboarding path="/onboarding" />
```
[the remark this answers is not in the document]
yjkimjunior: It works but I couldn't get the RouteComponentProps from there. But this bit is very much wip.

https://www.gatsbyjs.org/docs/porting-from-create-react-app-to-gatsby/#routing also suggests to use the Router for dynamic routes, so I went with that (also familiary structure as react-router)

front/gatsby/src/pages/onboarding.tsx
yjkimjunior: yes

front/context/src/AccountsContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { web3Accounts, web3Enable, isWeb3Injected } from '@polkadot/extension-dapp';
+import React, { useState, createContext, useEffect } from 'react';
+import { InjectedAccountExt } from './types';
+
+export const AccountsContext = createContext({
+  injectedAccounts: [] as InjectedAccountExt[]
+})
+
+interface Props {
+  children: any;
```
amaurymartiny: React.Children

front/context/src/AccountsContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function AccountsContextProvider (props: Props): React.ReactElement {
+  const { children } = props;
+  const [injectedAccounts, setInjected] = useState<InjectedAccountExt[]>([] as InjectedAccountExt[]);
```
amaurymartiny: Is the `as ...` necessary here?

front/gatsby/src/components/layout.tsx
```diff
@@ -5,14 +5,14 @@
...
+import { Container } from '@substrate/ui-components/src';
 import React from "react"
 import PropTypes from "prop-types"
 import { useStaticQuery, graphql } from "gatsby"
 
-import Header from "./header"
 import "./layout.css"
 
-const Layout = ({ children }: any) => {
+const Layout = ({ children }: any): React.ReactElement => {
```
amaurymartiny: Same, would avoid any everywhere

front/context/src/AccountsContext.tsx
yjkimjunior: maybe not the best way around it but the tsc error here was: `cannot assign never[] to InjectedAccountExt[]`

yjkimjunior:
> lgtm, we can merge like this. There are some linting issues, but we can fix them in a subsequent PR focused on adding a stricter linter.

sounds good to me, i'll lerna publish and then add the stricter eslint rules

PR #16, opened 2019-11-18
Feat: Onboarding with extension

yjkimjunior:
https://github.com/yjkimjunior/Nomidot/issues/3


yjkimjunior:
slight hiccup - the polkadotjs-extension doesn't have any setters, so we can't set the stash/controller intention that way. I'm thinking once we have the `db` we'll need to just set it there instead, but for now, i'll store an `'address': 'stash' | 'controller'` in localstorage as a workaround. What do you think? @amaurymartiny

amaurymartiny:
I added a small comment in https://github.com/yjkimjunior/Nomidot/issues/10#issuecomment-555498817.

So ideally, the fact that an account is stash or controller should be retrieved by DB1. We should never actually set a tag or something on an account (in our DB or in the extension) that indicates if it's stash or controller.

And for everything that's DB1, then i think we could directly use @polkadot/api for now. It'll be easier to transition once we have the DB, we would just need to replace `api.query.staking.bonded/ledger(someAddress) -> db.fetchControllerStash(someAddress)`

front/gatsby/src/components/profileCreationTutorial.tsx
```diff
@@ -0,0 +1,299 @@
...
+  };
+
+  // Step 3.
+  const renderConfirmMessage = (): React.ReactElement => {
+    return (
+      <WithSpaceAround>
+        <Stacked>
+          <Modal.SubHeader>
+            <DynamicSizeText fontSize='medium' fontWeight='500'>
+              Let &apos s get staking!
```
amaurymartiny:
```suggestion
              Let &apos;s get staking!
```

front/gatsby/src/components/profileCreationTutorial.tsx
```diff
@@ -0,0 +1,299 @@
...
+  InjectedAccountExt,
+  StakingContext,
+} from '@substrate/context/src';
+import {
+  AccountsList,
+  AddressSummary,
+  DynamicSizeText,
+  ErrorText,
+  Icon,
+  // Image,
```
amaurymartiny:
```suggestion
```

front/gatsby/src/components/profileCreationTutorial.tsx
```diff
@@ -0,0 +1,299 @@
+// Copyright 2018-2019 @paritytech/nomidot authors & contributors
```
amaurymartiny: Could you rename the file to have a capital first letter, and the same name as the component?

front/gatsby/src/pages/onboarding.tsx
```diff
@@ -1,45 +1,50 @@
...
+  }, []);
+
+  // TODO: save to DB not local storage
+  const handleSelectAccount = ({
+    currentTarget: {
+      dataset: { address },
+    },
+  }: React.MouseEvent<HTMLElement>): void => {
+    if (address) {
+      localStorage.setItem('exploringAs', address);
```
amaurymartiny: Could you add a `window.alert('TODO')` or something? While playing I wasn't sure if it was a bug or not implemented.

front/gatsby/src/pages/index.tsx
```diff
@@ -1,29 +1,40 @@
...
         }: Partial<ApiContextType>): React.ReactElement | boolean | undefined =>
-          isReady && (
+          isReady ? (
             <Layout>
               <SEO title='Home' />
-              <Onboarding />
+              {exploringAs ? <Profile /> : <Onboarding />}
             </Layout>
+          ) : (
+            <Loading active inline />
```
amaurymartiny: `<Loading active inline />` should go inside `ApiContextProvider` in the context gate, as the `loading` prop. And then you won't need the consumer here anymore.

amaurymartiny:
In reply to https://github.com/yjkimjunior/Nomidot/pull/16#issuecomment-555500380

After our meeting yesterday, I think we should avoid import `@polkadot/api` altogether, both in the frontend and in the backend. Only the node-watcher should import it.

For now, let's put `// TODO db` in the codebase, as you did.

front/gatsby/src/html.js
```diff
@@ -0,0 +1,40 @@
+import React from "react"
```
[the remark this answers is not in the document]
yjkimjunior: Gatsby generates the public folder on each build, so the `index.html` file always get reset to their default configuratoin, modifying this html.js file lets us persist some things. For tis case it was for the semantic ui css CDN, but in the future if we have any more CDNs this would be the place to add them. https://www.gatsbyjs.org/docs/custom-html/

amaurymartiny:
closing for now, let's also not use localstorage anymore

PR #17, opened 2019-11-18
chore: ad eslint-plugin-react-hooks

yjkimjunior:


yjkimjunior:
rules are already there in `eslint.rc` but the dep was handled by CRA before, so just adding it manually now

PR #27, opened 2019-12-03
Add validator info boilerplate

joepetrowski:
Added:
- Guide to add info to list.
- Example entry in JSON.
- `img` directory - feel free to restructure, not sure if there's a more standard format for where these go.


yjkimjunior:
lgtm

PR #29, opened 2019-12-03
refactor some code

amaurymartiny:
Some little changes, the createBlockNumber task passes (run `yarn start`).


back/node_watcher/datamodel.prisma
```diff
@@ -1,17 +1,20 @@
 type BlockNumber {
-  number: Int! @unique @id
+  id: ID! @id
+  number: Int! @unique
```
amaurymartiny: https://github.com/prisma/prisma/issues/3233 to use Int and @id, i think we need a sequence (didn't look much into it). For simplicity we can keep it like this, or we could use their sequence command

back/node_watcher/src/index.ts
```diff
@@ -2,27 +2,18 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { ApiPromise, WsProvider } from '@polkadot/api';
-import { TypeRegistry } from '@polkadot/types';
+import { logger } from '@polkadot/util';
 
-import { nodeWatcher } from './node-watcher';
-import nomidotTasks from './nomidot-tasks';
+import { nodeWatcher } from './nodeWatcher';
```
amaurymartiny: I put nodeWatcher into a separate file, so that it's importable once we build.

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -0,0 +1,84 @@
+// Copyright 2018-2019 @paritytech/nomidot authors & contributors
```
amaurymartiny: I renamed all files to use camel case (for consistency)

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Object to represent the current status of the incrementor
+ */
+interface State {
+  blockNumber: BlockNumber;
+  specVersion: u32;
+}
+
+async function incrementor(
```
amaurymartiny: Removed the generator, used a recursive function instead. it's imo a bit clearer to read

back/node_watcher/src/index.ts
```diff
@@ -2,27 +2,18 @@
...
-  const api = new ApiPromise({ provider, registry });
-
-  api.isReady
-    .then(() => {
-      nodeWatcher([...nomidotTasks], api);
-    })
-    .catch(e => {
-      console.error(e);
-    });
+  nodeWatcher([nomidotTasks[0]]).catch(e => {
```
yjkimjunior:
```suggestion
  nodeWatcher(nomidotTasks).catch(e => {
```

back/node_watcher/src/nodeWatcher.ts
yjkimjunior: hmm, I guess... I think we should avoid recursion when possible, and in the context of enumerating blocks, I think a generator is more intuitive. But I'm not super bothered in this case

back/node_watcher/datamodel.prisma
yjkimjunior: I have some uncommited changes to address this, blocked (as far as testing) by the workspace bug we discuess on Riot.

back/node_watcher/src/nodeWatcher.ts
amaurymartiny: > I think we should avoid recursion when possible

Why?

PR #30, opened 2019-12-03
Fix linting and test

amaurymartiny:
This fixes `yarn test` and `yarn lint`

(feel free not to merge if that breaks your base PR)

### DIFF
--- a/back/node_watcher/src/nomidot.spec.ts
+++ b/back/node_watcher/src/nomidot.spec.ts
@@ -3,7 +3,6 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { ApiPromise, WsProvider } from '@polkadot/api';
-// import { TypeRegistry } from '@polkadot/types';
 
 // import nomidotTasks from './nomidot-tasks';
 
@@ -16,14 +15,15 @@ describe('Nomidot Tasks', () => {
   beforeEach(
     async (done): Promise<void> => {
       api = await ApiPromise.create({ provider: new WsProvider(parityKusama) });
+
       done();
     }
   );
 
   it('should read a single task', async done => {
-    const intentions = await api.query.staking.validators();
+    const currentValidators = await api.query.session.validators();
 
-    console.error(JSON.parse(JSON.stringify(intentions)));
+    console.error(currentValidators.toJSON());
 
     done();
   });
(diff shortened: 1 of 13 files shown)


back/node_watcher/src/tasks/createBlockNumber.ts
```diff
@@ -37,7 +37,7 @@ const createBlockNumber: Task<NomidotBlock> = {
 
     const write = await prisma.createBlockNumber({
       number: blockNumber.toNumber(),
-      authoredBy: authoredBy.toString(),
+      authoredBy: authoredBy.toHex(),
```
amaurymartiny: Still to be discussed, but maybe saving public key is better?

back/node_watcher/src/nomidot.spec.ts
```diff
@@ -16,14 +15,15 @@ describe('Nomidot Tasks', () => {
...
     async (done): Promise<void> => {
       api = await ApiPromise.create({ provider: new WsProvider(parityKusama) });
+
       done();
     }
   );
 
   it('should read a single task', async done => {
-    const intentions = await api.query.staking.validators();
+    const currentValidators = await api.query.session.validators();
```
amaurymartiny: staking.validators is a linked map, was taking more than 5 seconds and test failed after that.

jest.config.js
```diff
@@ -2,6 +2,6 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-const config = require('@polkadot/dev/config/jest');
-
-module.exports = Object.assign({}, config);
\ No newline at end of file
+module.exports = {
```
amaurymartiny: The reason tests were failing for me, was that polkadot/dev is using babel, and here we're using tsc. ts-jest is a tsc wrapper around jest

amaurymartiny: We could switch to babel (main argument: faster compile times), but tsc is the official one and imo simpler to use

back/node_watcher/src/tasks/createBlockNumber.ts
amaurymartiny: Actually, I'm still hesitant. Addresses will be way easier to inspect

yjkimjunior: maybe we just be redundant and have two columns in there - (public_key, address)?

jest.config.js
yjkimjunior: let's stick with tsc

back/node_watcher/src/nomidot.spec.ts
yjkimjunior: ah, sweet, thank you

yjkimjunior:
thanks a ton!

PR #32, opened 2019-12-04
chore: Improve docker-compose

amaurymartiny:
No need to run `yarn build`, docker-compose automatically does `prisma deploy`.

So the flow is:
- docker-compose up (--build)
- yarn start


back/node_watcher/docker-compose.yml
```diff
@@ -1,15 +1,23 @@
-version: '3'
+version: '3.7'
 services:
-  # node_watcher:
-  #   image: kusama_node_watcher
-  #   restart: always
-  #   ports:
-  #     - "3000:3000"
+  main:
```
amaurymartiny: renamed to main, or else `node_watcher_node_watcher_1` was displayed, instead of now `node_watcher_main_1`

back/node_watcher/package.json
```diff
@@ -1,35 +1,31 @@
...
-  "keywords": [
-    "Prisma",
-    "Postgres",
-    "Polkadot"
-  ],
   "devDependencies": {
-    "@types/jest": "^24.0.23",
-    "jest": "^24.9.0",
-    "ts-jest": "^24.2.0"
+    "prisma": "^1.34.10",
```
amaurymartiny: I added prisma here so user doesn't need to globally install it.

So instead of `primsa <command>`, it'd be `yarn prisma <command>`

back/node_watcher/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:alpine3.10
+
+ENV PRISMA_ENDPOINT http://prisma:4466
```
yjkimjunior: did you mean `http://localhost:4466`, or does prisma here actually refer to the image?

amaurymartiny: Yes,in the virtualnetwork docker-compose creates, to access a container you put its image name

yjkimjunior: ah right 👍

PR #34, opened 2019-12-05
Sync from tip

yjkimjunior:


back/node_watcher/src/nodeWatcher.ts
```diff
@@ -19,14 +19,27 @@ async function incrementor(
...
-  //   l.warn(`last known best finalized: ${lastKnownBestFinalized}`);
-  // }, 5000);
+  setInterval(() => {
+    async function updateFinalized() {
+      lastKnownBestFinalized = await api.derive.chain.bestNumberFinalized();
+      l.warn(`last known best finalized: ${lastKnownBestFinalized}`);
+
+      if (blockIndex > lastKnownBestFinalized.toNumber()) {
+        blockIndex = lastKnownBestFinalized.toNumber();
+      }
```
amaurymartiny: I think you can remove the 3 lines above

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -19,14 +19,27 @@ async function incrementor(
...
+    }
+
+    updateFinalized();
+  }, 5000);
 
   while (true) {
+    if (blockIndex > lastKnownBestFinalized.toNumber()) {
+      blockIndex = lastKnownBestFinalized.toNumber();
```
amaurymartiny: Also remove this line

PR #39, opened 2019-12-16
K8s: Cloudsql proxy

yjkimjunior:


back/node_watcher/Dockerfile
```diff
@@ -1,15 +1,17 @@
...
-RUN chmod +x /opt/bin/wait-for.sh
+# ADD https://raw.githubusercontent.com/eficode/wait-for/master/wait-for /opt/bin/wait-for.sh
+# RUN chmod +x /opt/bin/wait-for.sh
 
 # Install dependencies
 RUN yarn
 # Create the database schema
-CMD ["/opt/bin/wait-for.sh", "prisma:4466", "--", "yarn", "prisma", "deploy"]
+# CMD ["/opt/bin/wait-for.sh", "prisma:4466", "--", "yarn", "prisma", "deploy"]
+CMD ["yarn", "prisma", "deploy"]
```
yjkimjunior: so currently a super annoying thing is that when this deploy script runs, it will try to open an oauth window, and that's not handled. Will look into how to fix that later.

PR #76, opened 2020-01-08
chore: Make CI green again

amaurymartiny:


front/context/src/AccountsContext.tsx
```diff
@@ -24,7 +23,12 @@ export function AccountsContextProvider(props: Props): React.ReactElement {
   useEffect(() => {
     const getInjected = async (): Promise<void> => {
       if (typeof window !== 'undefined') {
-        const { web3Accounts, web3Enable } = require('@polkadot/extension-dapp');
+        const {
```
yjkimjunior: it was breaking the ui as `@polkadot/extension-dapp` needs a global window to be defined when initialized. I agree this is a bit hacky but this seems to be the accepted workaround according to SO

PR #78, opened 2020-01-08
feat: Make ApiContext take a `provider` prop

amaurymartiny:
wait for #76 to be merged first


scripts/clean-build.sh
```diff
@@ -1,37 +0,0 @@
-
```
amaurymartiny: I removed this script, because there was an error in https://gitlab.parity.io/parity/nomidot/-/jobs/348609#L210, surely due to bash on linux and macos being not 100% the same.

rimraf is cross-platform

amaurymartiny:
I'll just merge it... seems like CI works only on master?

amaurymartiny:
oops, I merged it into am-ci. Will merge it into master again.

PR #86, opened 2020-01-10
feat: Create proposal and proposalArguments

Tbaut:
- adds a proposal table with unique proposal id (like on-chain)
- adds a proposalArguments table with a one-to-many relation with proposal so that I can do query like:
```gql
query {
  proposals {
    proposalId
    proposalArguments {
      name
      value
    }
  }
}
```

Tested with several proposal simultaneously:
- `system.setRemark` with 1 argument
- `balances.forceTransfer` with 3 arguments

Not sure how to best write a test for that with a public node :/


back/node_watcher/src/tasks/createValidator.ts
```diff
@@ -3,6 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { ApiPromise } from '@polkadot/api';
+import { Option } from '@polkadot/types';
```
Tbaut: Totally unrelated, all these changes where made by an auto fix on save :(

amaurymartiny:
I just merged master, let's check that CI is green

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -9,6 +9,8 @@ import { logger } from '@polkadot/util';
 import { NomidotTask } from './tasks/types';
 
 const ARCHIVE_NODE_ENDPOINT = 'wss://kusama-rpc.polkadot.io/';
+// const ARCHIVE_NODE_ENDPOINT = 'ws://127.0.0.1:9944';
```
amaurymartiny: I would prefer not to have commented code in the repo (exactly as in substrate repo).

Even a comment "Replace with 'ws://127.0.0.1:9944'" would do

amaurymartiny:
```suggestion
```

amaurymartiny:
It's green!

PR #94, opened 2020-01-19
Create proposal events and arguments

Tbaut:


Tbaut:
This is included in #96

PR #100, opened 2020-01-21
chore: Bump polkadot/api

Tbaut:


yjkimjunior:
bumped it in https://github.com/paritytech/Nomidot/pull/102, which also seems to have gotten the node watcher to start syncing again :)

PR #102, opened 2020-01-22
fix: k8s poststart hook, add back api to nodewatcher

yjkimjunior:
somehow this slipped.

### DIFF
--- a/back/node_watcher/kubernetes/staging/deployments/nomidotwatcher-deployment.yaml
+++ b/back/node_watcher/kubernetes/staging/deployments/nomidotwatcher-deployment.yaml
@@ -27,7 +27,4 @@ spec:
           env:
             - name: PRISMA_ENDPOINT
               value: http://10.0.9.43:4466
-          lifecycle:
-            postStart:
-              exec:
-                command: ["yarn", "start"]
+          command: ["yarn", "start"]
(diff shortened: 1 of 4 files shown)


back/node_watcher/package.json
```diff
@@ -24,6 +24,7 @@
     "start": "ts-node ./src/index.ts"
   },
   "dependencies": {
+    "@polkadot/api": "1.0.0-beta.15",
```
amaurymartiny: There are 2 instances in yarn lock

back/node_watcher/kubernetes/staging/deployments/nomidotwatcher-deployment.yaml
```diff
@@ -27,4 +27,7 @@ spec:
           env:
             - name: PRISMA_ENDPOINT
               value: http://10.0.9.43:4466
-          command: ["yarn", "start"]
+          lifecycle:
+            postStart:
+              exec:
+                command: ["yarn", "start"]
```
amaurymartiny: I'm not 100% sure what this means, but I'll trust you!

yjkimjunior: ah just tested it out but this is actually not necessary

Issue #101, opened 2020-01-22
bug(nodewatcher): update types for specVersion 1032


yjkimjunior:
pseudocode
```
if 1019 <= specversion < 1032:
injectTypes(types: {
      DispatchError: 'DispatchErrorTo198',
      Keys: 'SessionKeys5'
    })

if specversion >= 1032
  inject types ( types: {
      Keys: 'SessionKeys5'
    })
```

see https://github.com/polkadot-js/api/blob/master/packages/api/src/base/typeInjector.ts#L33

amaurymartiny:
Yeah, but don't hardcode anything (e.g. "1019", "DispatchErrorTo198") in Nomidot, just use the 2 exported functions from the link above

PR #104, opened 2020-01-23
fix(node-watcher): Override with chain types

Tbaut:
closes https://github.com/paritytech/Nomidot/issues/101


yarn.lock
```diff
@@ -2598,6 +2598,15 @@
     "@polkadot/api" "^1.0.0-beta.15"
     "@polkadot/types" "^1.0.0-beta.15"
 
+"@polkadot/api-derive@^1.0.0-beta.17":
```
amaurymartiny: Not sure why, but now there are 2 version of `@polkadot/api` in yarn.lock. By experience, this often causes hidden surprises

Tbaut:
I tried with `yarn add`, `yarn upgrade`.. and I always end up with an additional `polkadot/api` on top of the 15. Can't this happen if some other package require specifically this version ?

package.json
```diff
@@ -22,7 +22,7 @@
     ]
   },
   "dependencies": {
-    "@polkadot/api": "^1.0.0-beta.15"
+    "@polkadot/api": "1.0.0-beta.15"
```
Tbaut: this is also weird. Something's wrong.. it added `.18` to the `yarn.lock`, but not the package.json..?

Tbaut:
Reverted again. Feel free to try on your machine to see if you can get 1 `polkadot/api` version.

package.json
amaurymartiny: So there are two "@polkadot/api": one in the root package.json, another in back/node_watcher/package.json. You'd need to :
- either remove "@polkadot/api" from one of the 2 places (i'd say from the root one)
- or `yarn add` in both places

probably the 1st one is simpler, so that in the future we only have one place to update "@polkadot/api"

Tbaut: That makes sense. The front needs the package so I'll remove it from the back and upgrade again, that should solve it.

Tbaut:
Should be better now :)

PR #105, opened 2020-01-24
Create Referenda

Tbaut:
Note that if you get a weird error such as "internal server error" it needs eventually a `docker-compose down -v && prisma deploy` because of this `is Int! @sequence`

### DIFF
--- a/back/node_watcher/src/tasks/createPreimage.ts
+++ b/back/node_watcher/src/tasks/createPreimage.ts
@@ -149,11 +149,13 @@ const createPreimage: Task<NomidotPreimage[]> = {
           hash: h.toString(),
           metaDescription,
           method,
-          proposal: {
-            connect: {
-              id: p.id,
-            },
-          },
+          proposal: p
+            ? {
+                connect: {
+                  id: p.id,
+                },
+              }
+            : undefined,
           preimageArguments: {
             create: pA,
           },
@@ -167,11 +169,13 @@ const createPreimage: Task<NomidotPreimage[]> = {
               status,
             },
           },
-          referendum: {
-            connect: {
-              id: r.id,
-            },
-          },
+          referendum: r
+            ? {
+                connect: {
+                  id: r.id,
+                },
+              }
+            : undefined,
           section,
         });
       })
--- a/back/node_watcher/src/tasks/createPreimage.ts
+++ b/back/node_watcher/src/tasks/createPreimage.ts
@@ -56,7 +56,7 @@ const createPreimage: Task<NomidotPreimage[]> = {
           !preimageArgumentsRaw.AccountId
         ) {
           l.error(
-            `At least one of preimageArgumentsRaw: Hash, Balance or AccountId missing. ${preimageArgumentsRaw.Hash}, ${preimageArgumentsRaw.Balance}, ${preimageArgumentsRaw.AccountId}`
+            `At least one of preimageArgumentsRaw: Hash, Balance or AccountId missing: ${preimageArgumentsRaw.Hash}, ${preimageArgumentsRaw.Balance}, ${preimageArgumentsRaw.AccountId}`
           );
           return null;
         }
@@ -102,7 +102,7 @@ const createPreimage: Task<NomidotPreimage[]> = {
           author: preimageArguments.author,
           depositAmount: preimageArguments.depositAmount,
           hash: preimageArguments.hash,
-          metaDescription: meta?.documentation.toString(),
+          metaDescription: meta.documentation.toString(),
           method,
           preImageArguments,
           section,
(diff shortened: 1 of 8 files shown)


Tbaut:
Forgot to add the link between preimage and referendum. Doing it now.

Tbaut:
Should be good now.
Testing with blocks 92825 and then 100800, there's a proposal set, a preimage, then the proposal gets tabled and a referendum created.

back/node_watcher/src/tasks/createPreimage.ts
```diff
@@ -135,18 +135,30 @@ const createPreimage: Task<NomidotPreimage[]> = {
...
+        const r = referenda[0];
 
         await prisma.createPreimage({
           author: author.toString(),
-          depositAmount: depositAmount?.toString(),
+          depositAmount: depositAmount.toString(),
           hash: h.toString(),
+          metaDescription,
+          method,
+          proposal: p
```
yjkimjunior: question: if a preimage ends up not being connected to a proposal, i.e. `p` was undefined, what's supposed to happen? My guess is that shouldn't happen in the first place and you added the ` : undefined` to satisfy `tsc`?

back/node_watcher/src/tasks/createPreimage.ts
```diff
@@ -157,8 +169,13 @@ const createPreimage: Task<NomidotPreimage[]> = {
               status,
             },
           },
-          metaDescription,
-          method,
+          referendum: r
```
yjkimjunior: same question as above

back/node_watcher/src/tasks/createPreimage.ts
Tbaut: It actually happens in real life, because you "should" actually submit a preimage before submitting a proposal, so that ppl can see what you are proposing.
But you need a deposit for a preimage I think it will be dependant on the code you put onchain. So what happens is that the proposals for `setCode` for instance for a runtime upgrade are often submited after the proposal has been submitted because you don't want to have too much money locked in the deposit, if the referendum doesn't pass.

So in real life, proposal, referendum and preimage don't have a defined order. If you want your proposal to be executed (once it passed), it needs to have a preimage.

On the other side, back to your question, if you submit a preimage, and then never use it for a proposal, it uses storage onchain and there a mechanism so that it can be reaped (after a certain time I imagine, and the reaper will get the deposit back). I don't know the reaping process but there's a preimage event for this.

Tbaut:
TL;DR I added this proposal/referendum preimage connection in such a way that anything can happen in any order. Because live things have happened in any order.

back/node_watcher/src/tasks/createPreimage.ts
yjkimjunior: > So what happens is that the ~proposals~ (i guess you mean preimage here?) for setCode for instance for a runtime upgrade are often submited after the proposal has been submitted

so to be clear: you submit proposal, wait for it to go to referendum, then submit a preimage so you can be sure your deposit will locked for as little time as necessary?

also, in this case how can someone decide whether or not to second a proposal if there is no preimage?

also is that deposit decided by the size of the call or preimage, or is it a constant, or something else?

Tbaut: > (i guess you mean preimage here?)

Absolutely :)

> so to be clear: you submit proposal, wait for it to go to referendum, then submit a preimage so you can be sure your deposit will locked for as little time as necessary?

Yup, looks like it.
Also I saw Gav ask for someone else to submit the preimage for him because his internet connection was flaky.

> also, in this case how can someone decide whether or not to second a proposal if there is no preimage?

Since the preimage hash is deterministic, you can just check whatever the proposer tells you (off chain, e.g this is a setCode from this branch). The preimage hash is submitted with the proposal, it's just that the preimage itself isn't present there. so I agree it's an annoyance and I'm pretty sure it'll happen less and less in the future, but this has happened in Kusama before (didn't have the courage to check polkascan and find one).

> also is that deposit decided by the size of the call or preimage, or is it a constant, or something else?

Don't take my word for it, @joshua-mir told me afaicr but I might have misinterpreted it.

yjkimjunior: makes sense, thanks for the explanation!

Tbaut:
I did a quick check for something else and here are example:

- setCode, 1.3Mb: 130KSM deposit: [link](https://polkascan.io/pre/kusama-cc3/democracy/preimage/0x44e6c5bfef03f426fdc54fd157130e42915f74c484287bedc19ec6de63c0c6c8)
- Another one, setMaxMember, a couple cents deposit: [link](https://polkascan.io/pre/kusama-cc3/democracy/preimage/0x8e2ccc745d0fa65361d1f49c781887cf63af5a023563f69ff2445c39c6125fc5)

PR #107, opened 2020-01-27
Create referendum status

Tbaut:
- linked to https://github.com/paritytech/polkassembly/issues/218

I tried creating several proposals and referendum, making them pass, not pass, executed and things were logged accordingly.

### DIFF
--- a/back/node_watcher/src/tasks/createReferendumStatus.ts
+++ b/back/node_watcher/src/tasks/createReferendumStatus.ts
@@ -5,12 +5,12 @@
 import { ApiPromise } from '@polkadot/api';
 import { BlockNumber, Hash } from '@polkadot/types/interfaces';
 import { logger } from '@polkadot/util';
+import { referendumStatus } from 'src/util/statuses';
 
 import { prisma } from '../generated/prisma-client';
 import {
   NomidotReferendumRawEvent,
   NomidotReferendumStatusUpdate,
-  referendumStatus,
   Task,
 } from './types';
 
--- a/back/node_watcher/src/tasks/createReferendumStatus.ts
+++ b/back/node_watcher/src/tasks/createReferendumStatus.ts
@@ -5,9 +5,9 @@
 import { ApiPromise } from '@polkadot/api';
 import { BlockNumber, Hash } from '@polkadot/types/interfaces';
 import { logger } from '@polkadot/util';
-import { referendumStatus } from 'src/util/statuses';
 
 import { prisma } from '../generated/prisma-client';
+import { referendumStatus } from '../util/statuses';
 import {
   NomidotReferendumRawEvent,
   NomidotReferendumStatusUpdate,
--- a/back/node_watcher/src/tasks/createReferendumStatus.ts
+++ b/back/node_watcher/src/tasks/createReferendumStatus.ts
@@ -8,9 +8,9 @@ import { logger } from '@polkadot/util';
 
 import { prisma } from '../generated/prisma-client';
 import {
-  NomidotProposalRawEvent,
-  NomidotProposalStatusUpdate,
-  ProposalStatus,
+  NomidotReferendumRawEvent,
+  NomidotReferendumStatusUpdate,
+  referendumStatus,
   Task,
 } from './types';
 
@@ -19,27 +19,31 @@ const l = logger('Task: Referendum Status Update');
 /*
  *  ======= Table (Proposal Status Update) ======
  */
-const createProposal: Task<NomidotProposalStatusUpdate[]> = {
+const createProposal: Task<NomidotReferendumStatusUpdate[]> = {
   name: 'createProposalStatusUpdate',
   read: async (
     blockHash: Hash,
     api: ApiPromise
-  ): Promise<NomidotProposalStatusUpdate[]> => {
+  ): Promise<NomidotReferendumStatusUpdate[]> => {
     const events = await api.query.system.events.at(blockHash);
 
-    const proposalEvents = events.filter(
+    // The "Started" event is taken care of by the createReferendum
+    // task, so we need to filter it out.
+    const referendumEvents = events.filter(
       ({ event: { method, section } }) =>
-        section === 'democracy' && method === ProposalStatus.TABLED
+        section === 'democracy' &&
+        Object.values(referendumStatus)
+          .filter(status => status !== referendumStatus.STARTED)
+          .includes(method)
     );
 
-    const results: NomidotProposalStatusUpdate[] = [];
+    const results: NomidotReferendumStatusUpdate[] = [];
 
     await Promise.all(
-      proposalEvents.map(async ({ event: { data, typeDef } }) => {
-        const proposalRawEvent: NomidotProposalRawEvent = data.reduce(
+      referendumEvents.map(async ({ event: { data, typeDef, method } }) => {
+        const referendumRawEvent: NomidotReferendumRawEvent = data.reduce(
           (prev, curr, index) => {
             const type = typeDef[index].type;
-
             return {
               ...prev,
               [type]: curr.toString(),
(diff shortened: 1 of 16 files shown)


back/node_watcher/src/tasks/types.ts
```diff
@@ -123,13 +124,28 @@ export enum ProposalStatus {
...
-  STARTED = 'Started',
-}
+export const referendumStatus = {
+  CANCELLED: 'Cancelled',
+  EXECUTED: 'Executed',
+  NOTPASSED: 'NotPassed',
+  PASSED: 'Passed',
+  STARTED: 'Started',
+};
+
```
Tbaut: FIXME See if there is a nicer way to use types for this `ReferendumStatus`

back/node_watcher/src/tasks/types.ts
```diff
@@ -123,13 +124,28 @@ export enum ProposalStatus {
...
+// export type ReferendumStatus =
```
Tbaut: This is what I tried, but I often get annoyed by "type string can't be assigned to `ReferendumStatus`"

back/node_watcher/src/tasks/types.ts
```diff
@@ -123,13 +124,28 @@ export enum ProposalStatus {
...
+//   | 'Cancelled'
+//   | 'Executed'
+//   | 'NotPassed'
+//   | 'Passed'
+//   | 'Started';
+
```
Tbaut: I used to have an enum, but it doesn't allow to do the following because of type missmatch string != ReferendumStatus :
```js
Object.values(referendumStatus)
.filter(status => status !== referendumStatus.STARTED)
.includes(method)
```

back/node_watcher/src/tasks/types.ts
```diff
@@ -123,13 +124,28 @@ export enum ProposalStatus {
...
 }
 
-export enum ReferendumStatus {
-  CANCELLED = ' Cancelled',
-  EXECUTED = 'Executed',
-  NOTPASSED = 'NotPassed',
-  PASSED = 'Passed',
-  STARTED = 'Started',
-}
+export const referendumStatus = {
```
Tbaut: That's not a type.. is that ok to have this here?
Also because it's not a type, I can't use it above for instance in the `NomidotReferendum`, I have to use a string. 
Is there is a nicer way to do?

Tbaut:
@amaurymartiny @yjkimjunior 
I tried a number of things with this enum / object.
So far I think my favourite way to go is with enum, the only drawback is that you can't iterate over it, but at least you get a stronger type safety... Please check out my comments and let me know what you think.

back/node_watcher/src/tasks/types.ts
```diff
@@ -123,13 +124,28 @@ export enum ProposalStatus {
...
+//   | 'Passed'
+//   | 'Started';
+
+// export enum ReferendumStatus {
+//   CANCELLED = 'Cancelled',
+//   EXECUTED = 'Executed',
+//   NOTPASSED = 'NotPassed',
+//   PASSED = 'Passed',
+//   STARTED = 'Started',
+// }
```
Tbaut: I used to have an enum, but it doesn't allow to do the following because of type missmatch string != ReferendumStatus :
```js
Object.values(referendumStatus)
.filter(status => status !== referendumStatus.STARTED)
.includes(method)
```

back/node_watcher/src/tasks/types.ts
```diff
@@ -123,13 +124,28 @@ export enum ProposalStatus {
...
+  PASSED: 'Passed',
+  STARTED: 'Started',
+};
+
+// export type ReferendumStatus =
+//   | 'Cancelled'
+//   | 'Executed'
+//   | 'NotPassed'
+//   | 'Passed'
+//   | 'Started';
```
Tbaut: This is what I tried to keep using types, but I just like enum, I get annoyed by "type string can't be assigned to `ReferendumStatus`"

back/node_watcher/src/tasks/types.ts
```diff
@@ -123,13 +124,28 @@ export enum ProposalStatus {
...
-  PASSED = 'Passed',
-  STARTED = 'Started',
-}
+export const referendumStatus = {
+  CANCELLED: 'Cancelled',
+  EXECUTED: 'Executed',
+  NOTPASSED: 'NotPassed',
+  PASSED: 'Passed',
+  STARTED: 'Started',
+};
```
Tbaut: That's not a type.. is that ok to have this here?
Also because it's not a type, I can't use it above for instance in the `NomidotReferendum`, I have to use a string. 
Is there is a nicer way to do?

amaurymartiny: Yeah, this one is still alright i guess. What we also sometimes do, where there's shared logic, is put the shared logic in an `util/` folder, here `./util/referenda.ts`.

> I can't use it above for instance in the NomidotReferendum

You would need to use ReferendumStatus in NomidotReferendum.

The way you hardcoded `export type ReferendumStatus` is good. There is a pure typescript way ([here](https://www.typescriptlang.org/play/?ssl=10&ssc=1&pln=1&pc=1#code/KYDwDg9gTgLgBAYwgOwM7ysAZsTyAmArgLYDKMAhjIanALxwDeAUHG3AMICCAchwKIAZQfwAiALjgByDhWQJgAG0XB8UgDSt2-ABr8OAVQAqYyVP4hgCQjFUatbHgHkjABS6lSp6TwgxXFKiodprscO6e3lIBQSEOcKRGXABKJhLS5BSwcQC+cIGIKOgA3MzMMACeYMBwydi4wAQkmdS0DAAUldUQWHCYOHhEZJStAJQA2gDWwBU9cF3Ac-0NTcNUNAC6pUA)) but i'm not sure it's more readable, so it's okay to double hardcode them.

back/node_watcher/src/tasks/createReferendumStatus.ts
```diff
@@ -0,0 +1,109 @@
...
+    api: ApiPromise
+  ): Promise<NomidotReferendumStatusUpdate[]> => {
+    const events = await api.query.system.events.at(blockHash);
+
+    // The "Started" event is taken care of by the createReferendum
+    // task, so we need to filter it out.
+    const referendumEvents = events.filter(
+      ({ event: { method, section } }) =>
+        section === 'democracy' &&
+        Object.values(referendumStatus)
```
amaurymartiny: `Object.values(referendumStatus)` returns `string[]`. Here, we could cast it: `Object.values(referendumStatus) as ReferendumStatus[]`

amaurymartiny:
Gave some comments about using the object, but if you feel like enum is easier, i'm okay with that too

back/node_watcher/src/tasks/createReferendumStatus.ts
Tbaut: If I remember correctly, such a casting wasn't accepted, because the types "string" and `ReferendumStatus` have nothing in common according to TS.

I ended up using type of key of, which is not super readable at first, but I think it's nice.

back/node_watcher/src/tasks/types.ts
Tbaut: Thanks, I've put all the statuses in a util file.

Tbaut:
Thanks, I would really like to avoid duplicating things, and [your](https://www.typescriptlang.org/play/?ssl=10&ssc=1&pln=1&pc=1#code/KYDwDg9gTgLgBAYwgOwM7ysAZsTyAmArgLYDKMAhjIanALxwDeAUHG3AMICCAchwKIAZQfwAiALjgByDhWQJgAG0XB8UgDSt2-ABr8OAVQAqYyVP4hgCQjFUatbHgHkjABS6lSp6TwgxXFKiodprscO6e3lIBQSEOcKRGXABKJhLS5BSwcQC+cIGIKOgA3MzMMACeYMBwydi4wAQkmdS0DAAUldUQWHCYOHhEZJStAJQA2gDWwBU9cF3Ac-0NTcNUNAC6pUA) `key of type of` is doing exactly what is needed and prevents duplication, so I think it ok to trade of readability.

PR #110, opened 2020-01-27
fix: image and k8 for recent fixes

yjkimjunior:


yjkimjunior:
syncing again,,,

back/node_watcher/package.json
```diff
@@ -24,6 +24,7 @@
     "start": "ts-node ./src/index.ts"
   },
   "dependencies": {
+    "@polkadot/api": "^1.0.0-beta.18",
```
amaurymartiny: Thibaut removed it from this package, so that we only put it in the root one

yjkimjunior: yeah but when we dockerize it, the image will have no `@polkadot/api`

amaurymartiny: ah right

Issue #116, opened 2020-01-28
bug: nodewatcher pod restart with stateful block index


yjkimjunior:
when a pod is restarted, because the block index progress is not persisted, it will start over from the beginning.

this should not happen.

either the nodewatcher needs to be stateless or we need to put that blockindex somewhere else

Issue #109, opened 2020-01-27
nodewatcher: rolling update


yjkimjunior:
So after doing more research and getting more familiar with the tools, I think a good way to maintain this nodewatcher with rolling updates is as follows:

### Case 1. We want to track more things (no mutate existing entries)
we have up to 2 nodewatcher deployments: 1 curr, 1 next (ephemeral).

`curr` is the one that's running at the moment, closest to, or at, the tip of the chain.
`next` is only running if there is an update that needs to happen and fill in the blanks left by `curr`.

When want to track more things, i.e. add a table or column, we currently:
stop the deployment, and restart it from block 0, just to fill in the diff. The problem is that we end up no longer tracking the tip.

Instead, we can have a `next` deployment which does the above, leaving `curr` to do what it does, and filling in the blanks in behind it.

`next` can be updated to `curr` when they're both at the tip, and then `next` can be shut down until there is another update that needs to happen.

The `next`/`curr` stuff in practice would look like:

1. `kubectl create -f nomidot-watcher-curr.yaml` (e.g. v0.1)
2. `docker tag new-deployment-image && docker push new-deployment-image`
3. `kubectl create -f nomidot_watcher-next.yaml` (e.g. v0.2)
4. (after curr/new are synced): `kubectl rolling-update nomidotwatcher-curr --image=next-image`

### Case 2. We want to change how we track existing entries
In this case, currently we'd need to just purge the db and start over with the new image.

The better way would be similar to the above. We have a `curr` db, and a `next` db. 

We just sync the `next` db and leave any services pointing to `curr` db until it's synced, then switch over the `DB_NAME` in `nodewatcher-deloyment.yaml` so `next` becomes `curr` and `curr` can either be purged or kept.

PR #119, opened 2020-01-28
fix: nomidotwatcher pod should not restart

yjkimjunior:
fixes: https://github.com/paritytech/Nomidot/issues/116

weell it doens't really fix it, but it at least makes the problem a lot more easy to discover and recover from.


yjkimjunior:
actually this should be a Job not a Deployment

back/node_watcher/kubernetes/staging/nomidotwatcher-job.yaml
```diff
@@ -8,23 +8,20 @@ metadata:
...
+      restartPolicy: Never
       containers:
         - name: nomidotwatcher
-          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:73f32773eb4e397e9f4cc7028945750f8f1116c7e751179120300a18413797ee
+          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:23bda9bd84369d74bc4f9a299649f388e4676bf37700c82a2df2cb577808d00a
           env:
             - name: PRISMA_ENDPOINT
               value: http://10.0.9.43:4466
+            - name: START_FROM
+              value: '38000'
```
yjkimjunior: this way I don't have to create and push a whole new image everytime i want to recover from a failure from a particular branch. just need to do `kubectl replace -f ...`

back/node_watcher/kubernetes/staging/nomidotwatcher-job.yaml
```diff
@@ -1,5 +1,5 @@
-apiVersion: apps/v1
-kind: Deployment
+apiVersion: batch/v1
+kind: Job
```
yjkimjunior: this way it runs the script to completion, or tries to. if something fails then it won't strrt over and waste cpu

back/node_watcher/kubernetes/nomidotwatcher-job.yaml
```diff
@@ -1,30 +1,38 @@
...
       containers:
         - name: nomidotwatcher
-          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:73f32773eb4e397e9f4cc7028945750f8f1116c7e751179120300a18413797ee
+          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:22592d41b85880683b595de06f3b399eef3b42750baf02f6afc73e85130152ea
+          imagePullPolicy: Always
           env:
             - name: PRISMA_ENDPOINT
               value: http://10.0.9.43:4466
+            - name: START_FROM
+              value: '37921'
```
amaurymartiny: Are you trying to spawn multiple nomidot watchers? Why start at this block?

yjkimjunior: there was a bug with proposals around there, so process would exit, pod restarts from where the last job started (i had hardcoded for the previous version). It's a workaround just to get a full sync at least once so we have one to work with while figuring out a more sustainable way https://github.com/paritytech/Nomidot/issues/109

yjkimjunior:
closing to make a clean PR. too many things slipped in here already

PR #120, opened 2020-01-29
refactor(context): Split ApiContext into Api & SystemContext

amaurymartiny:
needed for https://github.com/paritytech/substrate-light-ui/pull/743

- Remove staking
- ApiContext only injects api
- SystemContext injects: chain properties, version, name, health, header etc, i.e. everything that doesn't need api
- KeyringContext now handles a `currentAccount` (related https://github.com/paritytech/substrate-light-ui/pull/415)


front/context/src/SystemContext.tsx
```diff
@@ -0,0 +1,152 @@
+// Copyright 2018-2020 @paritytech/substrate-light-ui authors & contributors
```
yjkimjunior: 🎉

front/gatsby/src/ContextGate.tsx
```diff
@@ -24,7 +23,7 @@ export function ContextGate({
       <AccountsContextProvider>
         <TxQueueContextProvider>
           <ApiContextProvider provider={new WsProvider(ARCHIVE_NODE_ENDPOINT)}>
-            <StakingContextProvider>{children}</StakingContextProvider>
+            <>{children}</>
```
yjkimjunior:
```suggestion
            {children}
```

front/ui-components/src/stateful/Balance.tsx
```diff
@@ -26,12 +26,12 @@ export function Balance(props: BalanceProps): React.ReactElement {
...
     ...rest
   } = props;
-  const { api, isReady } = useContext(ApiContext);
+  const { api, isApiReady } = useContext(ApiContext);
   const [allBalances, setAllBalances] = useState();
   const [allStaking, setAllStaking] = useState();
 
   useEffect(() => {
-    if (isReady) {
+    if (isApiReady) {
```
yjkimjunior: I find it weird that we have to do this everywhere, since none of these should load in the first place before the `isApiReady` is set to true in the ContextGate.

front/gatsby/src/ContextGate.tsx
amaurymartiny: Unfortunately, i get some error like https://github.com/paritytech/substrate-light-ui/pull/743/files#diff-418f37370656dbd0c860458c05c4933cR17

front/ui-components/src/stateful/Balance.tsx
amaurymartiny: > since none of these should load in the first place before the isApiReady is set to true in the ContextGate

Hmm, if we put this in ui-components and expose this package to npm, then I don't think we should assume that the end-user will user this component inside a ContextGate when isApiReady is set to true. Just to be sure, I'd leave this check here.

In light-ui, if we use internal components that are behind a Gate, then yeah we can remove it.

amaurymartiny: Edit: but also, as we talked before, a better way would simply only expose pure/stateless components in ui-components

front/gatsby/src/ContextGate.tsx
yjkimjunior: ah, i guess that makes sense....

PR #121, opened 2020-01-30
fix: make nodewatcher sync again

yjkimjunior:
went down many rabbitholes eventually found @polkadot/util-beta was what was killing me


yarn.lock
```diff
@@ -2612,6 +3647,20 @@
     "@polkadot/types" "^1.0.0-beta.18"
     "@polkadot/util-crypto" "^2.0.0-beta.5"
 
+"@polkadot/api@^1.0.0-beta.44":
```
amaurymartiny: twice this package again

yjkimjunior: i'll be more careful, haven't found a better way yet to manage these dependencies for dockerizing a single workspace

Issue #125, opened 2020-01-31
bug: nodewatcher create proposal status


yjkimjunior:
small bug crashed node watcher just now:

```
NODE-WATCHER: Task --- createProposalStatusUpdate
2020-01-31 01:08:20            MAIN: Error: Variable '$where' cannot be non input type 'ReferendumWhereUniqueInput!'. (line 1, column 16):
query ($where: ReferendumWhereUniqueInput!) {
               ^
    at BatchedGraphQLClient.<anonymous> (/node_watcher/node_modules/http-link-dataloader/src/BatchedGraphQLClient.ts:74:13)
    at step (/node_watcher/node_modules/http-link-dataloader/dist/src/BatchedGraphQLClient.js:40:23)
    at Object.next (/node_watcher/node_modules/http-link-dataloader/dist/src/BatchedGraphQLClient.js:21:53)
    at fulfilled (/node_watcher/node_modules/http-link-dataloader/dist/src/BatchedGraphQLClient.js:12:58)
    at processTicksAndRejections (internal/process/task_queues.js:88:5) {
  result: { data: null, errors: [ [Object], [Object], [Object] ], status: 200 }
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #126, opened 2020-02-03
fix:  #125 - createReferendumStatus

yjkimjunior:


yjkimjunior:
ugh, so it fixed locally but seems something didn't update when deployed, same error.

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -80,7 +80,6 @@ async function incrementor(
 
       try {
         l.warn(`Writing: ${JSON.stringify(result)}`);
-        await task.write(blockNumber, result);
```
Tbaut: Missed that one, is that intentional?

yjkimjunior:
> ugh, so it fixed locally but seems something didn't update when deployed, same error.

so what was happening is that `prisma deploy` was not actually updating the schema on the cluster. Added some docs for port forwarding which allows us to manually update it locally.

Issue #106, opened 2020-01-24
nodewatcher needs a recovery mechanism when Connection Dropped.


yjkimjunior:
Currently we need to update the container to start syncing from the block where connection was dropped, then restart the pods manually.

Ideally we'd have a way for this to happen automatically.

The error: 
` API-WS: disconnected from wss://kusama-rpc.polkadot.io/ code: '1006' reason: 'Connection dropped by remote peer.'`

yjkimjunior:
FWICT, `autoconnect` is default true on the api provider end so it does continue to try to reconnect...

whenever this happens in a browser, we just reload. 

when this happens in a pod we need to restart.

from what i've read, you just have to do it manually. https://stackoverflow.com/questions/46123457/restart-container-within-pod?rq=1

yjkimjunior:
Update:
ok so debugging this with Felix, what we're seeing is that Memory spikes around the same time when Stackdriver is logging that websocket connection refused, which is why the Pod gets Evicted rather than continuously trying to reconnect.

<img width="1173" alt="Screenshot 2020-02-10 at 12 44 01" src="https://user-images.githubusercontent.com/10432070/74147323-41c0f900-4c03-11ea-9be0-8094aab6002f.png">

So now we'll try to run a new job where we purposely set an incorrect endpoint and see if this memory spike happens then as well.

yjkimjunior:
currently trying: remove the backofflimit

Tbaut:
FWIW I just witnessed a pod printing API-WS: disconnected from wss://kusama-rpc.polkadot.io/ code: '1006' reason: 'connection failed' during 3.5min and then keep going when things went back to normal, so that's pretty cool.

Also it printed a couple:
```bash
2020-02-18 16:07:18    NODE-WATCHER: Writing: {"amount":"0x00000000000000007581be1bb86640cd"}
2020-02-18 16:07:18    NODE-WATCHER: FetchError: request to http://10.0.9.43:4466 failed, reason: connect ECONNREFUSED 10.0.9.43:4466
    at ClientRequest.<anonymous> (/node_watcher/node_modules/isomorphic-fetch/node_modules/node-fetch/index.js:133:11)
    at ClientRequest.emit (events.js:196:13)
    at ClientRequest.EventEmitter.emit (domain.js:471:20)
    at Socket.socketErrorListener (_http_client.js:402:9)
    at Socket.emit (events.js:196:13)
[...]

PR #129, opened 2020-02-03
fix (#106): try once to recreate api object if remote node drops connection

yjkimjunior:
#106 FWICT, as long as the socket remains open, @polkadot/api will be able to continue to retry the connection.

But sometimes the remote node will drop the connection (https://stackoverflow.com/questions/1434451/what-does-connection-reset-by-peer-mean), and it'll be up to us to restart the script back up from where the last block, which is quite tedious work. Hopefully this will be able to recover from those situations more often than not.


back/node_watcher/src/nodeWatcher.ts
```diff
@@ -33,12 +33,21 @@ function waitFinalized(
...
+  provider: WsProvider,
   tasks: NomidotTask[]
 ): Promise<void> {
   let blockIndex = parseInt(process.env.START_FROM || '0');
   let currentSpecVersion = api.createType('u32', -1);
   let lastKnownBestFinalized = await waitFinalized(api, 0);
 
+  api.once('disconnected', async () => {
+    try {
+      api = await ApiPromise.create({ provider });
```
amaurymartiny: ah, I don't like mutating function arguments. But i'm very curious if this works

yjkimjunior:
yeah I agree, though I pushed a job with this change just to test and it's been the longest running job without a bug so far. 

> more robust retry system (e.g. with exponential delay).

^ +1 for this as a longer term solution though

PR #136, opened 2020-02-04
feat(gatsby): Inject accounts into header

amaurymartiny:
- [x] Remove useless gatsby stuff (e.g. image optimization, useless static graphql queries)
- [x] Refactor injected accounts to use same types as `@polkadot/inject-dapp`
- [x] Add header and inject accounts
- [x] Update ui-keyring


front/gatsby/tsconfig.json
```diff
@@ -1,4 +1,10 @@
 {
+  "compilerOptions": {
+    "baseUrl": "../..",
+    "paths": {
+      "@substrate/context": ["front/context/src"]
```
amaurymartiny: works well!

front/context/src/AccountsContext.tsx
```diff
@@ -2,59 +2,71 @@
...
 
   return (
-    <AccountsContext.Provider value={{ injectedAccounts }}>
+    <AccountsContext.Provider
+      value={{
+        accounts,
+        get extension(): InjectedExtension {
+          if (!extension) {
+            throw new Error(
+              'Please use `extension` only after `isReady` is set to true'
```
amaurymartiny:
```suggestion
              'Please use `extension` only after `isExtensionReady` is set to true'
```

front/context/package.json
```diff
@@ -18,7 +18,7 @@
     "test": "echo skipped"
   },
   "dependencies": {
-    "@polkadot/ui-keyring": "^0.47.1",
+    "@polkadot/ui-keyring": "^0.48.1",
```
yjkimjunior: since we're on beta for the rest of the polkadot-js suite, this should also be `0.49.0-beta.1`. fine to do in a later PR though.

front/context/src/AccountsContext.tsx
```diff
@@ -2,59 +2,71 @@
...
-
-    getInjected();
-  }, []);
+    if (!extensions.length) {
+      throw new Error(
+        'No extension found. Please install PolkadotJS extension.'
+      );
+    }
+
+    setExtension(extensions[0]);
```
yjkimjunior: I'm fine with this for now before incorporate @goldsteinsveta 's designs for switching between accounts.

PR #145, opened 2020-02-06
fix(ui-components): Stateless InputAddress

amaurymartiny:
Remove it from the `stateful/` folder


amaurymartiny:
It was stateful because it relied on ui-keyring https://github.com/paritytech/Nomidot/pull/145/files#diff-cf99e41d9346cb3a0cbcc072d10e8347L6, which itself is bound to localStorage.

PR #146, opened 2020-02-06
feat: apollo client and provider

yjkimjunior:


front/gatsby/gatsby-browser.js
```diff
@@ -0,0 +1,45 @@
...
+);
+
+const client = new ApolloClient({
+  link,
+  cache: new InMemoryCache()
+})
+
+// eslint-disable-next-line react/prop-types
+export const wrapRootElement = ({element}) => (
+  <ApolloProvider client={client}>{element}</ApolloProvider>
```
amaurymartiny: I would actually be in favor of removing this file, because it's in JavaScript.

So i would:
- create an `apollo.ts`, exporting a React context provider
- in `pages/index.tsx`, do a `import { ThatProvider } from './path/to/apollo.ts'`
- put that provider next to `AccountsContextProvider`

yjkimjunior: yeah i like that better too, I just went with this during the attempts to solve the subscription bug before, seems this is the "gatsby way" of doing things fwir.

yjkimjunior:
just cutting this off at the bare minimum gluing, working with Sveta on the props for the design-systems components to work nicely with the way the data comes in now.

front/gatsby/gatsby-ssr.js
```diff
@@ -0,0 +1,5 @@
+// Copyright 2018-2020 @paritytech/Nomidot authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+// export { wrapRootElement } from './gatsby-browser';
```
amaurymartiny: Can delete this file i guess

front/gatsby/tsconfig.json
```diff
@@ -3,8 +3,9 @@
     "baseUrl": "../..",
     "paths": {
       "@substrate/context": ["front/context/src"]
-    }
+    },
+    "typeRoots": ["./design-system.d.ts"]
```
amaurymartiny: Could you try without this line? Somehow deleting it still make tsc work on my machine

front/gatsby/src/apollo.ts
```diff
@@ -0,0 +1,40 @@
...
+import { HttpLink } from 'apollo-link-http';
+import { WebSocketLink } from 'apollo-link-ws';
+import { getMainDefinition } from 'apollo-utilities';
+
+const httpLink = new HttpLink({
+  uri: 'https://34.76.106.193:4000',
+});
+
+const wsLink = new WebSocketLink({
+  uri: 'ws://34.76.106.193:4000',
```
yjkimjunior: btw this should probably be hidden or access controlled somehow?

yjkimjunior: @amaurymartiny bump

amaurymartiny: For production, yeah for sure.

For now, probably a devops could answer you better, i'm not sure how okay it's to expose public endpoints on the test installation on gcp.

yjkimjunior: cc: @fevo1971 @gabreal ?

gabreal: I guess I'm lacking some background here. is that portal going to be deployed on kubernetes? does it come with it's own polkadot client?

yjkimjunior: Soo we have a server as a deployment on kubernetes and a loadbalancer service for clients (such as this frontend) to connect to. But right now anyone with the above ExternalIP is able to connect.

 It does not come with its own polkadot client at the moment, though that might be helpful.

gabreal: so there is no authentication? - as long as it is temporary it okayish but if you intend to use it as a development platform it's probably better to think about access restriction.

PR #150, opened 2020-02-07
fix: referendum statuses not updated

Tbaut:
this time it's for referendum whose status was never updated.
Next time I'll test better!

### DIFF
--- a/back/node_watcher/src/tasks/createProposal.ts
+++ b/back/node_watcher/src/tasks/createProposal.ts
@@ -48,22 +48,28 @@ const createProposal: Task<NomidotProposal[]> = {
 
             return {
               ...prev,
-              [type]: curr.toString(),
+              [type]: curr.toJSON(),
             };
           },
           {}
         );
-
-        if (!proposalRawEvent.PropIndex || !proposalRawEvent.Balance) {
+        if (!proposalRawEvent.PropIndex && proposalRawEvent.PropIndex !== 0) {
           l.error(
-            `At least one of proposalArgumentRaw: PropIndex or Balance missing: ${proposalRawEvent.PropIndex}, ${proposalRawEvent.Balance}`
+            `Expected PropIndex missing on the event: ${proposalRawEvent.PropIndex}`
+          );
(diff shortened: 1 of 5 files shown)


back/node_watcher/src/tasks/createReferendumStatus.ts
```diff
@@ -63,7 +63,7 @@ const createReferendumStatus: Task<NomidotReferendumStatusUpdate[]> = {
         }
 
         const relatedReferendum = await prisma.referendum({
-          id: parseInt(referendumRawEvent.ReferendumIndex),
+          referendumId: parseInt(referendumRawEvent.ReferendumIndex),
```
amaurymartiny: Would it be possible to do `.toJSON()` [here](https://github.com/paritytech/Nomidot/blob/master/back/node_watcher/src/tasks/createReferendum.ts#L41), so that you would'nt need a parseInt

Tbaut: sure, let me try that.

Tbaut:
Here we go, it ended up being a little annoying because proposals and referenda can have the id 0, which is falsy.. but now at least there's no casting around.

back/node_watcher/src/tasks/createProposal.ts
```diff
@@ -48,22 +48,28 @@ const createProposal: Task<NomidotProposal[]> = {
 
             return {
               ...prev,
-              [type]: curr.toString(),
+              [type]: curr.toJSON(),
             };
           },
           {}
         );
+        if (!proposalRawEvent.PropIndex && proposalRawEvent.PropIndex !== 0) {
```
amaurymartiny: In this case `proposalRawEvent.PropIndex === undefined` (or `isUndefined` from `@polkadot/util`) might be easier to read

Tbaut: Much more readable indeed, but this leaves the `null` and empty string aside. I'd rather be extra safe and just check for this 0 exception.

Issue #135, opened 2020-02-04
Remove package-lock.json


amaurymartiny:
Let's use yarn everywhere.

https://github.com/paritytech/Nomidot/blob/master/back/node_watcher/package-lock.json

yjkimjunior:
oops not sure how that got in there

PR #151, opened 2020-02-08
chore(deps): Use kind-of >=6.0.3 everywhere

axelchalon:
This repo's dependencies currently use [older](https://github.com/paritytech/parity-signer/blob/97651dcc2f78027aa83c3dc7834f6941e084c8a7/yarn.lock#L5138) [versions](https://github.com/paritytech/Nomidot/blob/106a6879c95e5f1735e90def0bcce4e5b8df683e/back/node_watcher/package-lock.json#L7383) of kind-of which have a possibly exploitable vulnerability ([original issue](https://github.com/jonschlinkert/kind-of/issues/30), [PR](https://github.com/jonschlinkert/kind-of/pull/31)).

`node_watcher` still has the old `package-lock.json`; however I assume it could be removed? (as we use yarn everywhere + monorepo?)


yjkimjunior:
yeah that somehow snuck in, https://github.com/paritytech/Nomidot/issues/135, removing it here https://github.com/paritytech/Nomidot/pull/152

PR #153, opened 2020-02-10
chore(deps): bump api

yjkimjunior:


yarn.lock
```diff
@@ -3438,6 +3438,19 @@
     ip-regex "^4.1.0"
     moment "^2.24.0"
 
+"@polkadot/util@^2.2.1":
```
amaurymartiny: 2 utils

back/node_watcher/package.json
```diff
@@ -24,8 +24,8 @@
     "start": "ts-node ./src/index.ts"
   },
   "dependencies": {
-    "@polkadot/api": "^1.0.0-beta.44",
-    "@polkadot/util": "^2.0.0-beta.10",
+    "@polkadot/api": "^1.2.0-beta.4",
+    "@polkadot/util": "^2.2.1",
```
amaurymartiny: Can we remove this one actually? util is automatically included in api

yjkimjunior: it's mostly there for the logger FWIR, but I guess we just console.log to stdout anyway, just wont' have the nice colors

amaurymartiny: What I meant is that if you install `@polkadot/api`, then `@polkadot/util` will also be installed as a dependency. So we don't actually need this additional line (and it will also hopefully avoid having 2 of them in the future, since we'll only have the one bundled with `@polkadot/api`)

amaurymartiny:
ah f, sorry @yjkimjunior, I merged Axel's PR and it created conflicts in yours

Issue #111, opened 2020-01-27
Speed up Nodwatcher Sync


yjkimjunior:
Seems to average around 25 writes per second, which sounds good but that's approximately how many writes per block we have at this point. (there are only 86,400 seconds in a day)...

![Screenshot 2020-01-27 at 15 29 13](https://user-images.githubusercontent.com/10432070/73182746-e1698c00-4119-11ea-9a44-3bd2a1f33b49.png)

![Screenshot 2020-01-27 at 15 36 07](https://user-images.githubusercontent.com/10432070/73183231-c5b2b580-411a-11ea-8bf9-7ca446c9fb21.png)

The good news is that memory usage seems to be growing slower as time goes on, currently hovering around 1.75GB.

While I'm tempted to upgrade our instance to speed it up, the CPU utilization is only at like 5-10%, so I guess we should figure out how to squeeze more out of the current instance as it is.

yjkimjunior:
A pretty simple solution would be to have multiple scripts start from different blocks.

Say we have `N` scripts, and current best head `H`, we just run each script `i` from `block # ((H/N) * i)`.

Importantly, all but the last script deployment needs to know when to stop, so we avoid duplicate work. Probably just an environment variable per deployment will do.

yjkimjunior:
at the current speed though it'll need at least 10 workers in parallel to be able to sync in a couple days.

amaurymartiny:
I personally thinking spawning 10 workers is really overkill, and would rather understand what is actually taking so long before building the devops for 10 worksers.

So the question is: who's the bottleneck in those 25w/s:
- If the network requests between nodeWatcher and the archive node are taking 90% of that time, then maybe we should spawn an archive node in the same cluster.
- If the bottleneck is the nodeWatcher itself, maybe we should refactor the tasks. Right now we run those sequentially, but we could have easily 2-3x speedup by parallezing some tasks. E.g. define some dependency graph (e.g.: task1 needs task2 and task3, but all other tasks can run parallely).
- If the bottleneck is actually writing into Prisma, then maybe we should dig more into that direction?

Maybe we could refactor nodeWatcher to run some tasks in parallel?

yjkimjunior:
^^ sounds a lot better than having 10 workers.

I'm not entirely sure which is the main bottleneck, but the first 2 points are probably good to have in any case. I'll test out the second solution, if not seeing much improvment then try the first as well.

yjkimjunior:
[...]

PR #155, opened 2020-02-10
Added cluster to speed up block processing

niklabh:
Blocks are being processed from process.env.START_FROM to lastKnownBestFinalized

Blocks are processed one by one which takes time. We can leverage node js multi core processing with cluster module. cluster produces identical copy of running code with seperate stack, local variables and event loop. If number of cpu is 8 the block processing time will be reduced to 1/8 of the original. This code assume that block processing can be done out of order and in parallel.

Master and workers can be differentiated with cluster.isMaster cluster.isWorker flags.

I am dividing the blocks from start to end in equal batches of size end-start/numberOfCpus.
Processing different batch with formula

```
const batch = Math.floor((lastKnownBestFinalized - blockIndex) / cpuCount);
blockIndex = blockIndex + (cluster.worker.id - 1) * batch;
endAt = blockIndex + batch - 1;
```

In worker if am processing blockIndex to endAt. Then i am exiting the loop.

In master i am keeping the loop alive with getting waitFinalized each time as old logic. 

This code is not 100% tested and and should not be merged yet. Some corner cases are not covered. Same block could be processed multiple times.

Need to add code where you only want 1 process and not master worker as we dont want that during development. 

Please discuss and merge only after it is tested and cover all the corner cases.


amaurymartiny:
> This code assume that block processing can be done out of order and in parallel.

This assumption does not hold. Example:

- Cluster 1: Blocks 1000-2000:
  - Block 1500: A new proposal gets created, with id `id_a`
- Cluster 2: Blocks 2000-30000:
  - Block 2000: Proposal `id_a`'s status gets updated

There's a high probablibility that cluster2 will process block 2000 before cluster1 processes block 1500. In that case, cluster2's task will not find proposal `id_a` in the DB.

niklabh:
ah ok then closing

amaurymartiny:
I think a more promising avenue we could explore is what @yjkimjunior  proposed: do an ETL directly from an archive node's key-value DB into prisma's SQL db. This would allow easily a 1000x speedup, since we bypass all network requests done at each block.

The tricky part is how to find a way to continuosly read from the archive node's key-value DB while it syncs.

Basically we now have 3 ways of doing it, ordered from the most elegant to least elegant, aka also from most work to least work on our side:
1. Wait for substrate-archive to be done, do an ETL from substrate-archive's DB to our chain DB
2. ETL from archive-node's kvdb to our chain DB
3. What we do currently: do a bunch of block-scoped tasks, increment block, write into our chain DB.

I don't want to depend on 1, it might take 2 weeks or 2 months. If 3 continues to give us problems, we could consider exploring 2. wdyt @niklabh @yjkimjunior?

yjkimjunior:
I've never tried a RocksDB to Postgresql migration, but I dont imagine it's trivial especially with the way Substrate storage works. Given the resources we have at our disposal, I'd say we just hammer through with 3 until we are able to ship our respective projects, and if 1 is stilll not ready by then, then we revisit 2. 

3 is painful and manual right now but after we've synced the core parts I don't think the updates and resyncing will be as frequent. 

In terms of speed of syncing, I'd like to revisit the ideas of:

1. throw more CPUs at it
2. run more jobs in parallel from differnt START_FROM's

Both are pretty "dumb" from an engineering perspective, but would get the job done faster from a product perspective IMHO.

amaurymartiny:
Some other optimizations:
- cache some `.at()`, i see multiple `api.query.system.events.at()`, they always return the same result.
- run tasks (not jobs) parallely, see https://github.com/paritytech/Nomidot/issues/111#issuecomment-578882139

> run more jobs in parallel from differnt START_FROM's

Wouldn't this also possibly run into inconsistencies like https://github.com/paritytech/Nomidot/pull/155#issuecomment-584164004?

yjkimjunior:
> Wouldn't this also possibly run into inconsistencies like #155 (comment)?

ah 💩 yeah that's right

> cache some .at(), i see multiple api.query.system.events.at(), they always return the same result.

working on a PR that includes something like this, not caching but rather just checks if a new session started then only query the thing (eg rewards, validators, heartbeats only happen once a session).

> run tasks (not jobs) parallely

this one sounds in theory great but I'm not actually sure what a concrete implementation would look like. Would we group tasks at each level of the dependency graph into their own job and run those in parallel?

Issue #118, opened 2020-01-28
Build Polkassenbly front-server


Tbaut:
Not sure how to do this with prisma but I guess we can have something like an admin password and use this in node-watcher.

Right now anyone can change anything in the DB by just ~~knowing the IP~~ edit: accessing it from the same cluster.

yjkimjunior:
the cloudsql instance for the chain db has an admin user and password whose credentials are stored as a k8s pod secret.

Tbaut:
But because it's in the same cloud instance, our hasura stitching server can update/delete/insert anything. We can't secure that from hasura's end (yet, that's a feature in progress).

yjkimjunior:
ah yeah, that's true, still not anyone can write anything by just knowing the IP, they'd need to write from either a nomidot server or a polkassembly server, so probably the biggest concern is us shooting ourselves in the foot. I guess we could give our servers their own service account with a viewer role for the db and then you'd only be able to read even if the server is compromised

Tbaut:
But Nomidot will speak to the graphql server (that will speak to the DB with credentials) isn't it?
Well, because Nomidot is accessible from the outside, we can actually use any query as long as there's no restriction put in place.

> I guess we could give our servers their own service account with a viewer role for the db and then you'd only be able to read even if the server is compromised

The thing is that we don't talk to the DB (at least I don't), we talk to prisma/graphql server, that has all the rights. So I guess we need to see how to restrict the access to some queries.

yjkimjunior:
> But Nomidot will speak to the graphql server (that will speak to the DB with credentials) isn't it?

yes, but that doesn't make it "accessible from the outside" (if I'm understanding you correctly). Nomidot and its microservices are only accessible from within the same cluster. The DB exists separately outside of the cluster, which is why we need to use the proxy to create a secure connection.

The only attack vector I could see is if someone were to be able to ssh into a running prisma server container and then try to write from there. But at that point that means they've got a hold of our GCP credentials anyway.

Beyond the frontend being accessible from the outside, the only thing we need to enforce is that no servers have resolvers to write anything to chain db in the first place. We can add IAM roles to more strictly enforce that.

> So I guess we need to see how to restrict the access to some queries.

I'm thinking Queries (Reads), all access. Writes, zero access. All servers connecting to chaindb will have read-only permissions. Only the nodewatcher gets to write.

Tbaut:
[...]

PR #156, opened 2020-02-10
feat(server): Polkassembly chain-db server

Tbaut:
closes https://github.com/paritytech/Nomidot/issues/118
closes https://github.com/paritytech/polkassembly/issues/290


yjkimjunior:
should this be in Polkassembly?

Tbaut:
To me, it sits right in front of Prisma. I don't mind moving it elsewhere, but I really think it's a good idea to have your "server" (effectively nomidot front server) close to this one. But you decide.

Tbaut:
How did you get around the no explicit any @yjkimjunior ?

yjkimjunior:
> How did you get around the no explicit any lint errors @yjkimjunior ?

could just create a type? I think for some of this boilerplate stuff, e.g. `parent` it's all typed as `any` by Apollo themselves so I wouldn't waste too much time with that.

Tbaut:
> hmm that makes sense too. I would personally prefer it to be in Polkassembly just for separation of concerns but if this is easier for you, I don't mind too much for now

Sounds good, then let's keep this here for now and we can still move it any time it it disturbs, it's a matter of copy/pasting.

> could just create a type? I think for some of this boilerplate stuff, e.g. parent it's all typed as any by Apollo themselves so I wouldn't waste too much time with that.

I thought the CI wouldn't let us merge this, but it looks like it actually will. So all good.

back/polkassembly-front-server/kubernetes/server-namespace.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Namespace
```
yjkimjunior: don't think this needs to be duplicated here since it's the same namespace.

Tbaut: Not sure I get that. do you mean the same namespace as the `server`?

yjkimjunior: you already have a namespace `polkassembly` created from the Polkassembly repo, so why duplicate it here? ... unless you were trying to create another namespace for the server in which case the below lines should be edited.

Tbaut: Understood thanks, I've no clue what I'm doing when touching kubernetes config files.

Tbaut:
Added motions and tried some nested queries in vain unfortunately. I need to dig a little deeper into this.

yjkimjunior:
I'm fine to merge this here now if it unblocks you guys @niklabh @Tbaut

Tbaut:
Sorry to let this lag behind. 
My idea is to merge this with `server` in the next days, and to move to using it.

It's not extremely urgent since only parity ppl use Polkassembly now, but it'll get to the top soon.

Tbaut:
Here we go, I tested a nested query successfully. Unless I missed anything, this should cover all cases.

We'll move to it as soon as this is done https://github.com/paritytech/polkassembly/issues/372

Issue #141, opened 2020-02-05
nodewatcher: additional Tasks for Nomidot


yjkimjunior:
Ok forgot to log this before but need to add the following tasks:

* HeartBeat
* ImOnline
* Rewards
* Nominations
* Stakes

PR #157, opened 2020-02-10
feat(#141): heartbeats, offline, nominations, rewards, stakes

yjkimjunior:
https://github.com/paritytech/Nomidot/issues/141

- [x] HeartBeat
- [x] Offline Validators
- [x] Rewards
- [x] Nominations
- [x] Stakes

--
more things that Id like to do before merging:
- [x] - merge nominations and validators tasks into a single task
~- [x] - dont write stake and total issuance on a block unless it changed from previous block.~ edit: actually this is necessary
- [x] - dont use derive for the total stake.


back/node_watcher/src/tasks/createOfflineValidator.ts
```diff
@@ -0,0 +1,80 @@
...
+        sessionIndex: {
+          connect: {
+            index: sessionIndex.toNumber()
+          }
+        },
+        wasThereAnOffenceThisSession,
+        validatorId: validatorId.toString(),
+        total: total.toHex(),
+        own: own.toHex(),
+        others
```
yjkimjunior: grumble

back/node_watcher/src/tasks/index.ts
```diff
@@ -6,25 +6,36 @@ import createBlockNumber from './createBlockNumber';
...
   createEra,
   createSlashing,
   createTotalIssuance,
-  createValidator,
+  createNominationAndValidators,
+  createPreimage,
+  createProposal,
+  createProposalStatus,
+  createReferendum,
+  createReferendumStatus,
```
Tbaut: These are duplicated afaics. Weird that the CI didn't catch that.

yjkimjunior: 🙏🙏🙏 holy cow that would have sucked if you hadn't caught that

PR #159, opened 2020-02-10
feat(node-watcher): Store motions in chain-db

### LINKED ISSUE #148: [Node watcher] Track Council motions

Tbaut:
- closes #148


back/node_watcher/src/tasks/createMotion.ts
```diff
@@ -0,0 +1,165 @@
...
+import { prisma } from '../generated/prisma-client';
+import { motionStatus } from '../util/statuses';
+import { NomidotMotion, NomidotMotionRawEvent, Task } from './types';
+
+const l = logger('Task: Motions');
+
+/*
+ *  ======= Table (Motion) ======
+ */
+const createMotion: Task<NomidotMotion[]> = {
```
yjkimjunior: ah wait btw, these are PolkassemblyMotion! can go do a replacement in another PR to rename all the Polkassembly stuff as such though.

Tbaut: Sure. tbh I'd rather name that all "NomidotBlabla" because we do share things already such as blockbnumbrer, and the future might add some more..

yjkimjunior: in that case then we should remove Nomidot prefix entirely imo

Tbaut: Agree, let's use another PR for that though :smile:

PR #161, opened 2020-02-11
fix: Remove useless stuff from context & ui-components

amaurymartiny:
- remove label from InputAddress
- move validate function to light-ui, because those validations are UI-specific
- flush txQueue on tx close


front/context/src/TxQueueContext.tsx
```diff
@@ -88,14 +88,30 @@ export function TxQueueContextProvider(props: Props): React.ReactElement {
...
+
   /**
    * Unsubscribe the tx with id `extrinsicId`
    */
   const closeTxSubscription = (extrinsicId: number): void => {
     const tx = txQueue.find(tx => tx.id === extrinsicId);
     if (tx) {
       tx.unsubscribe();
-      setTxQueue(txQueue.splice(txQueue.indexOf(tx), 1));
+      clear(); // Clear all the txqueue once one if finalized
```
yjkimjunior: I'm not clear why this is the desired behavior?

Issue #160, opened 2020-02-11
enhancment: minimize queries


yjkimjunior:
Two places off the top of my head where we don't need to be querying as much as we are.

1. system.events.at(blockhash) -> currently querying per task. Easily can just put that in the main nodewatcher part.
2. things that happen once a session -> currently querying every block anyway, can easily just check whether a new session happened (or whether that session exists in the db), and then get on with it.

edit: in fact for 2, we are even querying the session index for each task. this can be pulled out higher like 1.

PR #169, opened 2020-02-13
feat(#160) - minimize queries

yjkimjunior:
#160 
- [x] : use filterEvents util function
- [x] : pull out event queries to once per block
- [x] : pull out session index queries to once per block
- [x] : make sure things that happen once per session only get queried when an entry doesn't already exist for that session
  - heartbeats: this comes from events so to query chaindb to check existence would actually be counterprodcutive.
  - nominators: done
  - validators: done

### DIFF
--- a/back/node_watcher/src/nodeWatcher.ts
+++ b/back/node_watcher/src/nodeWatcher.ts
@@ -83,8 +83,8 @@ async function incrementor(
 
     const [events, sessionIndex] = await Promise.all([
       await api.query.system.events.at(blockHash),
-      await api.query.session.currentIndex.at(blockHash)
-    ])
+      await api.query.session.currentIndex.at(blockHash),
+    ]);
 
     const cached: Cached = {
       events,
--- a/back/node_watcher/src/nodeWatcher.ts
+++ b/back/node_watcher/src/nodeWatcher.ts
@@ -81,11 +81,13 @@ async function incrementor(
       api.registry.setMetadata(rpcMeta);
     }
 
+    const events = await api.query.system.events.at(blockHash);
+
     // execute watcher tasks
     for await (const task of tasks) {
       l.warn(`Task --- ${task.name}`);
 
-      const result = await task.read(blockHash, api);
+      const result = await task.read(blockHash, events, api);
 
       try {
         l.warn(`Writing: ${JSON.stringify(result)}`);
--- a/back/node_watcher/src/nodeWatcher.ts
+++ b/back/node_watcher/src/nodeWatcher.ts
@@ -81,8 +81,10 @@ async function incrementor(
       api.registry.setMetadata(rpcMeta);
     }
 
-    const events = await api.query.system.events.at(blockHash);
-    const sessionIndex = await api.query.session.currentIndex.at(blockHash);
+    const [events, sessionIndex] = await Promise.all([
+      await api.query.system.events.at(blockHash),
+      await api.query.session.currentIndex.at(blockHash)
+    ])
 
     const cached: Cached = {
       events,
--- a/back/node_watcher/src/nodeWatcher.ts
+++ b/back/node_watcher/src/nodeWatcher.ts
@@ -7,7 +7,7 @@ import { BlockNumber, Hash } from '@polkadot/types/interfaces';
 import { getChainTypes } from '@polkadot/types/known';
 import { logger } from '@polkadot/util';
 
-import { NomidotTask } from './tasks/types';
+import { Cached, NomidotTask } from './tasks/types';
 
 const ARCHIVE_NODE_ENDPOINT =
   process.env.ARCHIVE_NODE_ENDPOINT || 'wss://kusama-rpc.polkadot.io/';
@@ -84,11 +84,16 @@ async function incrementor(
     const events = await api.query.system.events.at(blockHash);
     const sessionIndex = await api.query.session.currentIndex.at(blockHash);
 
+    const cached: Cached = {
+      events,
+      sessionIndex,
+    };
+
     // execute watcher tasks
     for await (const task of tasks) {
       l.warn(`Task --- ${task.name}`);
 
-      const result = await task.read(blockHash, events, sessionIndex, api);
+      const result = await task.read(blockHash, cached, api);
 
       try {
         l.warn(`Writing: ${JSON.stringify(result)}`);
--- a/back/node_watcher/src/nodeWatcher.ts
+++ b/back/node_watcher/src/nodeWatcher.ts
@@ -82,12 +82,13 @@ async function incrementor(
     }
 
     const events = await api.query.system.events.at(blockHash);
+    const sessionIndex = await api.query.session.currentIndex.at(blockHash);
 
     // execute watcher tasks
     for await (const task of tasks) {
       l.warn(`Task --- ${task.name}`);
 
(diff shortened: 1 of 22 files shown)


back/node_watcher/src/tasks/createBlockNumber.ts
```diff
@@ -17,7 +17,7 @@ const l = logger('Task: BlockNumber');
  */
 const createBlockNumber: Task<NomidotBlock> = {
   name: 'createBlockNumber',
-  read: async (blockHash: Hash, api: ApiPromise): Promise<NomidotBlock> => {
+  read: async (blockHash: Hash, events: EventRecord[], api: ApiPromise): Promise<NomidotBlock> => {
```
amaurymartiny: btw, we started with `Task` having `read` and `write`. When i have a look at the tasks, it seems that read/write could be merged together, the distinction is not always relevant

yjkimjunior: hmm,  I think for now the separation is a helpful way of organizing the code at least. But merging might actually be necessary anyway for the task parallelization PR down the road. I'd leave this for another PR

back/node_watcher/src/tasks/createHeartbeat.ts
```diff
@@ -19,10 +19,10 @@ const createHeartBeat: Task<NomidotHeartBeat[]> = {
   name: 'createHeartBeat',
   read: async (
```
yjkimjunior: @amaurymartiny how bad is it that now these arguments will be left unused in a lot of these tasks?

amaurymartiny: Performace wise: there would be no difference. These args are passed by reference (aka they are not cloned), so there's close-to-0 memory overhead.

amaurymartiny: Linting wise: yeah, we could add a rule that says if the variable is preceded by `_`, then it can be unused.

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -81,11 +81,14 @@ async function incrementor(
...
 
+    const events = await api.query.system.events.at(blockHash);
+    const sessionIndex = await api.query.session.currentIndex.at(blockHash);
+
     // execute watcher tasks
     for await (const task of tasks) {
       l.warn(`Task --- ${task.name}`);
 
-      const result = await task.read(blockHash, api);
+      const result = await task.read(blockHash, events, sessionIndex, api);
```
amaurymartiny: Could also consider to put these 2 args in a `cached: {events, sessionIndex}` object, so that you don't end up with 7 arguments soon

back/node_watcher/src/tasks/types.ts
```diff
@@ -30,7 +30,12 @@ import {
 
 export interface Task<T> {
   name: string;
-  read(blockHash: Hash, api: ApiPromise): Promise<T>;
+  read(
+    blockHash: Hash,
+    events: EventRecord[],
+    sessionIndex: SessionIndex,
+    api: ApiPromise
+  ): T | Promise<T>;
```
amaurymartiny: In `createOfflineValidator`, I'd actually rather have at the end `return Promise.resolve(result)`, so we keep a consistent `Promise<T>` interface

back/node_watcher/src/nodeWatcher.ts
yjkimjunior: will do

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -81,11 +81,19 @@ async function incrementor(
       api.registry.setMetadata(rpcMeta);
     }
 
+    const events = await api.query.system.events.at(blockHash);
+    const sessionIndex = await api.query.session.currentIndex.at(blockHash);
```
amaurymartiny: optimization inside optimization: could use Promise.all!

back/node_watcher/src/tasks/createStake.ts
```diff
@@ -18,7 +18,11 @@ const l = logger('Task: Stake');
...
   name: 'createStake',
-  read: async (blockHash: Hash, api: ApiPromise): Promise<NomidotStake> => {
+  read: async (
+    blockHash: Hash,
+    _cached: Cached,
+    api: ApiPromise
+  ): Promise<NomidotStake> => {
     const currentElected = await api.query.staking.currentElected.at(blockHash);
     const stakersInfoForEachCurrentElectedValidator: Exposure[] = [];
     let totalStaked = new BN(0);
```
amaurymartiny: unrelated to this PR: i would advocate to never use BN directly, always use `api.createType`

back/node_watcher/src/nodeWatcher.ts
amaurymartiny: wait, even api.queryMulti?

Tbaut:
ah I got a TS error actually:
```bash
TSError: ⨯ Unable to compile TypeScript:
src/tasks/createOfflineValidator.ts:27:3 - error TS2322: Type '(_blockHash: Hash, cached: Cached, _api: ApiPromise) => NomidotOfflineValidator[]' is not assignable to type '(blockHash: Hash, cached: Cached, api: ApiPromise) => Promise<NomidotOfflineValidator[]>'.
  Type 'NomidotOfflineValidator[]' is missing the following properties from type 'Promise<NomidotOfflineValidator[]>': then, catch, [Symbol.toStringTag], finally

27   read: (
     ~~~~

  src/tasks/types.ts:33:3
    33   read(blockHash: Hash, cached: Cached, api: ApiPromise): Promise<T>;
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The expected type comes from property 'read' which is declared here on type 'Task<NomidotOfflineValidator[]>'
```

back/node_watcher/src/nodeWatcher.ts
yjkimjunior: hmm I wonder which one of those would be better actually

amaurymartiny: queryMulti: it sends one WS message. Promise.all send 2.

yjkimjunior: ah cool 👌

Tbaut:
:+1: working well

back/node_watcher/src/nodeWatcher.ts
yjkimjunior: how do you queryMulti on `.at` decorated functions? this is erroring for me:
```
    api.queryMulti([
      [api.query.system.events.at, blockHash],
      [api.query.session.currentIndex.at, blockHash]
    ], ([events, sessionIndex]) => {
      
    })
```

but this works
```
    api.queryMulti([
      api.query.system.events,
      api.query.session.currentIndex
    ], ([events, index]) => {})
```

yjkimjunior: FWICT `StorageEntryPromiseOverloads` does not expect the decorations

amaurymartiny: ah, my bad, you cannot do queryMulti and .at, see https://github.com/polkadot-js/api/issues/1039

So Promise.all it is

Issue #173, opened 2020-02-14
UnhandledPromiseRejectionWarning: Error: Option: unwrapping a None value


Tbaut:
The last 50k job halted at:
```bash
2020-02-13 22:10:59    NODE-WATCHER: Task --- createEra
2020-02-13 22:10:59       TASK: ERA: NomidotEra: {"idx":411,"points":{"total":53440,"individual":[340,260,500,240,380,300,520,460,300,300,220,380,260,460,420,400,520,240,400,320,360,520,380,500,340,80,400,500,220,140,340,460,320,460,380,300,380,400,300,460,400,460,240,500,420,440,200,380,320,340,340,360,400,80,320,260,160,520,380,120,280,420,280,380,340,360,360,460,320,500,360,360,300,360,340,400,460,0,400,420,260,440,120,380,420,240,540,400,360,420,560,340,0,360,280,400,300,340,440,140,420,240,320,320,140,300,140,540,420,320,160,260,260,280,340,460,360,260,400,380,200,300,440,240,400,300,260,260,20,300,540,440,60,340,360,440,320,260,440,480,300,520,240,80,300,340,300,240,0,300,320,340,400,320,360,380,200,380,180,280]},"startSessionIndex":1803}
2020-02-13 22:10:59    NODE-WATCHER: Writing: {"idx":411,"points":{"total":53440,"individual":[340,260,500,240,380,300,520,460,300,300,220,380,260,460,420,400,520,240,400,320,360,520,380,500,340,80,400,500,220,140,340,460,320,460,380,300,380,400,300,460,400,460,240,500,420,440,200,380,320,340,340,360,400,80,320,260,160,520,380,120,280,420,280,380,340,360,360,460,320,500,360,360,300,360,340,400,460,0,400,420,260,440,120,380,420,240,540,400,360,420,560,340,0,360,280,400,300,340,440,140,420,240,320,320,140,300,140,540,420,320,160,260,260,280,340,460,360,260,400,380,200,300,440,240,400,300,260,260,20,300,540,440,60,340,360,440,320,260,440,480,300,520,240,80,300,340,300,240,0,300,320,340,400,320,360,380,200,380,180,280]},"startSessionIndex":1803}
2020-02-13 22:10:59    NODE-WATCHER: Task --- createSlashing
2020-02-13 22:10:59  TASK: SLASHING: Nomidot Slashing: []
2020-02-13 22:10:59    NODE-WATCHER: Writing: []
[...]

PR #174, opened 2020-02-14
fix(#173): unwrap none

yjkimjunior:
#173


back/node_watcher/src/generated/prisma-client/prisma-schema.ts
```diff
@@ -1,4 +1,4 @@
-// Code generated by Prisma (prisma@1.34.10). DO NOT EDIT.
+// Code generated by Prisma (prisma@1.23.0-test.3). DO NOT EDIT.
```
Tbaut: Any reason to downgrade?

yjkimjunior: no this was a mistake on my part

Issue #175, opened 2020-02-15
[Node-watcher] Kubernetes deployment fail - Unable to compile TypeScript 


Tbaut:
```bash
kubectl logs nomidotwatcher-1548a46-964699-dvt85 -n nodewatcher-staging --tail=30 --follow
yarn run v1.15.2
$ ts-node ./src/index.ts

/node_watcher/node_modules/ts-node/src/index.ts:421
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
src/nodeWatcher.ts(24,18): error TS2339: Property 'toNumber' does not exist on type 'BlockNumber'.
src/nodeWatcher.ts(25,24): error TS2339: Property 'toNumber' does not exist on type 'BlockNumber'.
src/nodeWatcher.ts(72,24): error TS2339: Property 'gt' does not exist on type 'U32'.

    at createTSError (/node_watcher/node_modules/ts-node/src/index.ts:421:12)
    at reportTSError (/node_watcher/node_modules/ts-node/src/index.ts:425:19)
    at getOutput (/node_watcher/node_modules/ts-node/src/index.ts:530:36)
    at Object.compile (/node_watcher/node_modules/ts-node/src/index.ts:735:32)
    at Module.m._compile (/node_watcher/node_modules/ts-node/src/index.ts:814:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:770:10)
    at Object.require.extensions.<computed> [as .ts] (/node_watcher/node_modules/ts-node/src/index.ts:817:12)
    at Module.load (internal/modules/cjs/loader.js:628:32)
    at Function.Module._load (internal/modules/cjs/loader.js:555:12)
    at Module.require (internal/modules/cjs/loader.js:666:19)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #176, opened 2020-02-15
Added types for bn.js. Other docker misc fixes

niklabh:
- types for bn.js
- dockerfile changes so that yarn only runs when dependencies has changed
- a yarn.lock file in node_watcher. please move root node.js project in seperate folder inside repo
- .dockerignore file so copy command dont copy node_modules


Tbaut:
closes: https://github.com/paritytech/Nomidot/issues/175

yjkimjunior:
> a yarn.lock file in node_watcher. please move root node.js project in seperate folder inside repo

how was this lock file generated? @niklabh

amaurymartiny:
Follow-up questions:
- why is it actually needed?
- could we just copy the one in the root folder inside the docker image?

niklabh:
Generated by deleting yarn.lock in root folder of the repo then running yarn in node_watcher

niklabh:
> Follow-up questions:
> 
> * why is this lock file actually needed? could we just copy the one in the root folder inside the docker image?

File need to be copied inside docker to make the builds fast. 
yarn.lock in root should not be copied. First of all this is not possible in docker as dockerfile works only in current working directory. Also yarn.lock in root is different from the node_watcher folder as package.json is differenet. When yarn.lock is present in parent folder that is used. This will conflict with root folder project and node_watcher project.

amaurymartiny:
Two things:
- I just saw https://github.com/polkadot-js/api/blob/master/CHANGELOG.md#131-feb-18-2020 (1st entry), I think types/bn.js could be removed now from package.json
- I still don't see how copying yarn.lock makes the build faster, but we can keep it. In this case, I would recommend to remove `back/node_watcher` from yarn workspaces

niklabh:
Build is fast by using this dockerfile

FROM node:10

RUN mkdir -p /opt/app

RUN chown node:node /opt/app

WORKDIR /opt/app

# yarn.lock should be in node_watcher to be copied into docker
COPY --chown=node:node package.json .
COPY --chown=node:node yarn.lock .

USER node
# docker work on layered architecture. if there is no change in above lines .i.e. package.json and yarn.lock this will not run again making build super fast
RUN yarn

COPY --chown=node:node . .

CMD yarn start

PR #178, opened 2020-02-17
feat: header ui with design-system

yjkimjunior:
cc: @goldsteinsveta 

The StatsList component is a bit difficult to use for this case, but as you can see I think the StatItem component works just fine for this header use case.


yjkimjunior:
pretty insubstantial at the moment so just going to merge to unblock myself.

.gitlab-ci.yml
```diff
@@ -20,6 +20,6 @@ linux-test:
     - yarn install --frozen-lockfile
     - yarn build
     - yarn test
-    - yarn lint
+    - yarn lint --fix
```
amaurymartiny: I recommend against that, imagine the following scenario:
- Small lint error on your local branch, you commit & push to a PR
- Gitlab run this command, fixes your small lint error on CI
- Gitlab **does not** push this lint fix to your branch or on master
- master has a small lint error

amaurymartiny: Would recommend:
- `yarn lint` on CI
- if CI does not pass, run `yarn lint --fix` on your local machine, then push again.
- CI should pass

yjkimjunior: ah, i didn't realize the 3rd point would be the case.

Issue #172, opened 2020-02-13
back/server: more subscriptition resolvers


yjkimjunior:
add subscription resolvers for all the new things that went into node_watcher

PR #180, opened 2020-02-17
feat (#172): subscription resolvers for stake, heartbeats, offline

yjkimjunior:
#172


yjkimjunior:
pretty standard boilerplate, merging to unblock myself

PR #182, opened 2020-02-17
feat: decoding SCALE hex, query then subscribe, Validators List

yjkimjunior:
Addresses #4 

- [x] : should decode SCALE hex
- [x] : subscription resolvers are listening for new field creations or updates so if we're listening for a new era, it should query the latest first, fill that in the frontened, then subscribe.
- [x] : some refactoring
- [x] : show current elected list

I'm also using some stuff from ui-components because not everything is ready with design-system yet (to be fair there a LOT of components that need to be built out). Personally I dont care as much that we mix and match components so long as we're moving forward quickly in a maintainable way.

Screenshot 2020-02-18 at 20 52 45


yjkimjunior:
https://recordit.co/5vCfc52PiT

front/gatsby/src/components/Layout/Header/Header.tsx
```diff
@@ -2,85 +2,71 @@
...
+        data: { eras },
+      } = queryData;
+
+      setEraHead({
+        index: eras[0].index,
+        individualPoints: api.createType(
+          'Vec<Points>',
+          eras[0].individualPoints
+        ),
+        totalPoints: api.createType('Points', eras[0].totalPoints),
```
amaurymartiny: I think we're fine for now. But it _might_ happen that this throws an error at some point, since we could be decoding past data with latest metadata/types.

front/gatsby/src/pages/index.tsx
```diff
@@ -13,11 +17,15 @@ import { APP_SLUG } from '../util';
...
   return (
     <ApolloProvider client={client}>
-      <AccountsContextProvider originName={APP_SLUG}>
-        <Layout>
-          <Seo title='Home' />
-        </Layout>
-      </AccountsContextProvider>
+      <ApiContextProvider
+        provider={new WsProvider('wss://cc3-5.kusama.network/')}
+      >
```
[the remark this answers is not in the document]
yjkimjunior: > api is only needed to do api.createType, correct?

At the moment, yes, but the thinking was that we will want it pretty soon to be able to nominate, bond, unbond, etc. directly from the app.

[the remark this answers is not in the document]
amaurymartiny: yeah, makes sense

front/gatsby/src/components/Layout/Header/Header.tsx
yjkimjunior: we could _maybe_ do something like you suggested for the other potential problem above + what we do in `node_watcher` such as:

"this is if my case for using the api above is valid"
```
export function createType(...args: AnyJSON) {
    // check spec version
    const runtimeVersion = await api.rpc.state.getRuntimeVersion(blockHash);
    const newSpecVersion = runtimeVersion.specVersion;

    // if spec version was bumped, update metadata in api registry
    if (newSpecVersion.gt(currentSpecVersion)) {
      l.warn(`bumped spec version to ${newSpecVersion}, fetching new metadata`);
      const rpcMeta = await api.rpc.state.getMetadata(blockHash);
      currentSpecVersion = newSpecVersion;

      // based on the node spec & chain, inject specific type overrides
      const chain = await api.rpc.system.chain();
      api.registry.register(getChainTypes(chain, runtimeVersion));

      api.registry.setMetadata(rpcMeta);
    }

    return api.createType(...args);
}
```

Pros:
* addresses the potential for a case where we'd need decode from an older version of metadata/types

Cons:
* it would mean 1-3 extra RPC calls per graphql subscription result (though this would be pretty infrequent, iff we show a different UI for when the DB has fallen behind the tip).

front/gatsby/src/components/Layout/Header/Header.tsx
```diff
@@ -159,17 +163,17 @@ const StakingHeader = () => {
       if (!stakeHead || stakeHead.blockNumber > number) {
         setStakeHead({
           blockNumber: number,
-          totalStake,
+          totalStake: api.createType('Balance', totalStake),
```
yjkimjunior: @amaurymartiny true, but https://github.com/paritytech/Nomidot/pull/169#discussion_r378940317 was still fresh in my mind 😅

front/gatsby/src/components/Layout/Header/Header.tsx
amaurymartiny: What about never storing anything SCALE-encoded in the DB, so that on the frontend, we never need to do any `createType` (aka no SCALE-decoding on the front)?

yjkimjunior: sounds good 👍(for the next minor version)

yjkimjunior:
since this is still under 500 LOC, i'll just go ahead fill in those missing columns in this PR as well

front/gatsby/src/components/Layout/Header/Header.module.css
```diff
@@ -1,5 +1,12 @@
-.header {
+.headerTop {
   display: flex;
   flex-direction: row;
-  justify-content: space-between;
+  justify-content: space-evenly;
+  align-items: stretch;
```
amaurymartiny: btw, that's the "official" way by Gatsby in their tutorial to add CSS. But a more elegant alternative could be using tachyon classes and removing all these .module.css files

we could have a chat with @goldsteinsveta to decide how we would like the design-system's components to look like, for the end user.

front/gatsby/src/components/Layout/Validators/CurrentElected.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <Table.HeaderCell>Controller</Table.HeaderCell>
+            <Table.HeaderCell>Commission</Table.HeaderCell>
+            <Table.HeaderCell> </Table.HeaderCell>
+          </Table.Row>
+        </Table.Header>
+        <Table.Body>
+          {currentElected ? (
+            currentElected.map(
+              ({ stash, controller, preferences, wasOfflineThisSession }) => (
+                <Table.Row textAlign='center' key={shortid.generate()}>
```
amaurymartiny: key is used for caching and optimization while re-rendering, adding this effectively removes all that.

I think `key={stash}` could be good? stash should be unique by row.

front/gatsby/src/components/Layout/Header/Header.tsx
```diff
@@ -2,91 +2,78 @@
...
-      totalStake
-    }
-  }
-`;
-
-const EraHeader = () => {
+const EraHeader = (): React.ReactElement => {
+  const { api } = useContext(ApiContext);
   const { data } = useSubscription(ERAS_SUBSCRIPTION);
+  const queryData = useQuery(LATEST_ERA_QUERY);
```
amaurymartiny: I'm curious: why do we need both?

front/gatsby/package.json
```diff
@@ -28,5 +29,8 @@
     "start:prod": "yarn build && yarn serve",
     "serve": "gatsby serve",
     "test": "echo \"Write tests! -> https://gatsby.dev/unit-testing\" && exit 1"
+  },
+  "devDependencies": {
+    "@types/shortid": "^0.0.29"
```
amaurymartiny: Makes more sense in general to add the `@types/*` next to the original one

front/gatsby/src/components/Layout/Validators/CurrentElected.tsx
yjkimjunior: up till recently that's what i would have done, however reading this suggests to use `shortid` https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

front/gatsby/src/components/Layout/Header/Header.tsx
yjkimjunior: the subscription only listens for a mutation, i.e. a new era was created. So if we just subscribe, this field will be empty till the next era is created. 

An alternative way to the above is to use polling queries but I think for something that doesn't change very often such as eras, the current implementation makes more sense.

front/gatsby/src/components/Layout/Header/Header.module.css
yjkimjunior: SGTM, I'm always up for less CSS.

front/gatsby/src/components/Layout/Validators/CurrentElected.tsx
yjkimjunior: upon further reading though, you're right, this is also an antipattern...sigh. will use stashid instead.

yjkimjunior: shite, so there are duplicate stashes. that should not be possible...

yjkimjunior: ok so i'll keep shortid for now, #190 should make sure to definitively remove any duplicates.

PR #183, opened 2020-02-18
Added Treasury in chain_db and node_watcher

niklabh:
- [x] Add treasury model in prisma
- [x] generate prisma bindings for treasury
- [x] Create task to read and write treasury events
- [x] Link treasury to council motions


back/node_watcher/datamodel.prisma
```diff
@@ -172,4 +172,21 @@ type ReferendumStatus {
...
\ No newline at end of file
+}
+
+type Treasury {
+  id: ID! @id
+  proposer: String!
+  beneficiary: String!
+  value: String!
+  bond: String!
+  proposalId: Int! @unique
```
Tbaut: I'd call that `treasuryProposalId` to mirror the `motionProposalId`, and leave `proposalId` to public proposals only.

back/node_watcher/datamodel.prisma
```diff
@@ -172,4 +172,21 @@ type ReferendumStatus {
   blockNumber: BlockNumber! @relation(name: "ReferendumStatusAtBlockNumber", link: INLINE)
   referendum: Referendum! @relation(name: "ReferendumStatusFromReferendum", link: INLINE)
   status: String!
-}
\ No newline at end of file
+}
+
+type Treasury {
+  id: ID! @id
```
Tbaut: You must have an `Int` here otherwise hasura will throw. I'd advise to carefully follow the same path as what I've done for proposal and referendum.

back/node_watcher/src/tasks/createTreasury.ts
```diff
@@ -0,0 +1,114 @@
...
+          treasuryRawEvent.ProposalIndex !== 0
+        ) {
+          l.error(
+            `Expected ProposalIndex missing in the event: ${treasuryRawEvent.ProposalIndex}`
+          );
+          return null;
+        }
+
+        const treasuryProposalRaw = await api.query.treasury.proposals(treasuryRawEvent.ProposalIndex);
+        const treasuryProposal: TreasuryProposal = api.createType('TreasuryProposal', treasuryProposalRaw.toU8a(true));
```
Tbaut: I think this `.toU8a(true)` is a mistake from my side. I have no clue what it really helps for, this came from some code that I found on `derive`. @yjkimjunior  or @amaurymartiny  can you shed some light on that, do you know if we should use it or not?

back/node_watcher/src/tasks/createTreasury.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        const proposalArguments: NomidotTreasuryEvent = {
+          proposalId: treasuryRawEvent.ProposalIndex,
+        };
```
Tbaut: Not sure what those lines help for. I'd use the `treasuryRawEvent.ProposalIndex` right away.
I might have done something similar in another task to make things type safe, but that might be just over-complicating things.

back/node_watcher/src/tasks/createTreasury.ts
```diff
@@ -0,0 +1,114 @@
...
+          beneficiary,
+          value,
+          bond,
+          proposalId,
+          status,
+        } = prop;
+
+        await prisma.createTreasury({
+          proposer: proposer.toString(),
+          proposalId,
```
Tbaut: reflecting my other comment. I don't mind the `proposalId` in this file, but for more clarity the db should have `treasuryProposalId`

back/node_watcher/src/tasks/types.ts
```diff
@@ -137,6 +138,22 @@ export interface NomidotProposalRawEvent {
...
+  proposer: AccountId;
+  beneficiary: AccountId;
+  value: Balance;
+  bond: Balance;
+  status: string;
+}
+
+export interface NomidotTreasuryEvent {
+  proposalId: number;
+}
```
Tbaut: most probably not needed.

back/node_watcher/src/util/statuses.ts
```diff
@@ -30,3 +30,8 @@ export const motionStatus = {
   PROPOSED: 'Proposed',
   VOTED: 'Voted',
 };
+
+export const treasuryProposalStatus = {
+  PROPOSED: 'Proposed',
+  TABLED: 'Tabled',
```
Tbaut: `Tabled` isn't a treasury event. What you'll care about for the status task is `Awarded` and `Rejected`.

back/node_watcher/datamodel.prisma
niklabh: done

back/node_watcher/datamodel.prisma
niklabh: done

back/node_watcher/src/tasks/createTreasury.ts
niklabh: replaced with unwrap. Previously api.query.treasury.proposals was returning a Codec now its returning Option<TreasuryProposal>. wierd

back/node_watcher/src/tasks/createTreasury.ts
niklabh: thanks, removed. It was unnecessary

back/node_watcher/src/tasks/createTreasury.ts
niklabh: changed to treasuryProposalId

back/node_watcher/src/tasks/types.ts
niklabh: removed

back/node_watcher/src/util/statuses.ts
niklabh: done

yjkimjunior:
@niklabh do you want to include "Link treasury to council motions" as well in this PR?

niklabh:
I will tackle that in different PR

yjkimjunior:
Ow actually, I think that will end up blocking the sync. i.e. We merge this and start syncing with Treasury not linked to council motions, then we get that other PR with the link, then we'll need to sync yet another DB.

Tbaut:
Agreed, it's just a couple lines of code to do a connect, it should be done in this pr

yjkimjunior:
bump @niklabh , would really like to have this one merged before starting v1.5 sync!

back/node_watcher/datamodel.prisma
```diff
@@ -180,3 +180,20 @@ type BlockIndex {
   startFrom: Int!
   index: Int!
 }
+
+type Treasury {
```
yjkimjunior: this should probably be named `TreasurySpendProposal` to avoid confusion

niklabh: but what else is in treasury other than spen proposals

yjkimjunior: well the treasury would have an avaialable balance, and spend period, for example

yjkimjunior: also this struct has a treasuryProposalId, which the Treasury in general wouldn't have

niklabh: done

niklabh:
> bump @niklabh , would really like to have this one merged before starting v1.5 sync!

Council motion is linked.

Tbaut:
What about my other comment? Did we test this?
From what I can see, nothing will ever be linked because the createMotion task will only get triggered for council section and not treasury. But I might very well have missed something in my non thorow review.

niklabh:
I have tested using local by sending spend proposal to council. createMotion task is being triggered

Tbaut:
Nice, I miss read that, the section comes from the arguments indeed. I'd still prefer to have this logic together with the one responsible for proposals in the read part, so that we have all the custom motion logic at the same place, I'll ease future understanding of the code I think. But it's not a must for now, this can be adapted later.

niklabh:
> Nice, I miss read that, the section comes from the arguments indeed. I'd still prefer to have this logic together with the one responsible for proposals in the read part, so that we have all the custom motion logic at the same place, I'll ease future understanding of the code I think. But it's not a must for now, this can be adapted later.

Can we move in read part in different PR. YJ needed this to be merged now.

Issue #184, opened 2020-02-18
Store processed blockIndex in persistant storage in node_watcher.


niklabh:
- processed blockIndex should be stored in persistant storage so that if pod get killed it can pickup from last processed block.

PR #185, opened 2020-02-18
Added persistance blockIndex

niklabh:
Closes: https://github.com/paritytech/Nomidot/issues/184
- added blockIndex in blockIndex table
- added IDENTIFIER so multiple runs can be done in parallel


back/node_watcher/src/util/blockIndex.ts
```diff
@@ -0,0 +1,30 @@
+import { BlockIndexCreateInput, prisma } from '../generated/prisma-client'
+
+export function createBlockIndex(identifier: string, startFrom: number, index: number) {
```
Tbaut: I guess all these functions here could be used directly.
There is a stray `a` function and `updateBlockIndex` that is empty.

niklabh: removed

niklabh:
Unrelated lint is failing

yjkimjunior:
could you just run `yarn lint --fix` locally and then push that?

niklabh:
issue is in nomidot/back/server shouldn't it go in seperate PR.

niklabh:
also lint command is not there in server or node_watcher

yjkimjunior:
you can run it from either root or from /back, CI says it's a couple trailing commas so i think it's trivial enough to just fit into this PR

PR #191, opened 2020-02-19
feat: app wide routing setup with @reach/router

yjkimjunior:
doing things the gatsby way took a while to figure out...

started (then commented out) the work in AccountsContext for tagging everything in the keyring as stash or controller or neither. It feels really ugly to do things this way but since the alternative is to get the user to input it and get it right, this seems to be the better way forward.


front/gatsby/package.json
```diff
@@ -21,6 +25,7 @@
     "react": "^16.11.0",
     "react-dom": "^16.11.0",
     "react-helmet": "^5.2.1",
+    "react-transition-group": "^4.3.0",
```
amaurymartiny: is this used somewhere?

front/gatsby/gatsby-browser.js
```diff
@@ -0,0 +1,38 @@
...
+    </ApiContextProvider>
+  </ApolloProvider>
+);
+
+export const wrapPageElement = ({ element }) => (
+  <Layout>
+    <Seo title='Polkadot/Kusama Staking Portal' />
+    <GlobalStyle />
+    {element}
+  </Layout>
```
amaurymartiny: Having a `wrapRootElement` is still okay, because I think this rootElement will really be the same on all pages

However, `wrapPageElement`'s SEO might be different on each page, e.g. if we want a validator's name to appear in the title.

amaurymartiny: A TS-only alternative would be:

```tsx
export function RootElement() {
  <AllTheContexts>
    <Layout, SEO, GlobalStyle />
  </AllTheContext>
}

// in pages/something.tsx
export function MyPage() {
  <RootElement seoTitle='MyPage'>
    {page-specific-UI}
  <RootElemetn>
}
```

yjkimjunior: `wrapPageElement` can receive props data about the location (and others from RouteComponentProps) from `@reach/router` so we can later pass down the appropriate SEO based on that. From what I've read, `wrapRootElement` is the Gatsby idiomatic place to put most if not all Providers, and `wrapPageElement` the place to put wrapper components that won’t get unmounted on page changes.

1. perhaps we can work around it with: https://github.com/gatsbyjs/gatsby/issues/1457#issuecomment-381405638
2.  I don't think that it is in itself a reason to go the long winded non-Gatsby way, considering the selling point of Gatsby is its magical wonders.

front/gatsby/package.json
yjkimjunior: no (not yet),

front/gatsby/gatsby-browser.js
yjkimjunior: docs for reference: https://www.gatsbyjs.org/docs/browser-apis/#wrapPageElement

Issue #192, opened 2020-02-20
"Where" arguments aren't passed down the parent queries


Tbaut:
I just realized that any query with `where`, basically ignores those (silently :sob:) when pointing the query at the "front" server. It's working fine on 4466.

Example:
```gql
query getSessions {
  sessions (where: {index_gt: 3}){
    index
  }
}
```

returns:
```json
{
  "data": {
    "sessions": [
      {
        "index": 1
      },
      {
        "index": 2
      },
      {
        "index": 3
      },
      {
        "index": 4
      }
    ]
  }
}
```

where 4466 returns:
```json
{
  "data": {
    "sessions": [
      {
        "index": 4
      }
    ]
  }
}
```

PR #194, opened 2020-02-20
Fix where args not taken into account

Tbaut:
closes https://github.com/paritytech/Nomidot/issues/192

That was a weird one. I basically change any query from 
```ts
  sessions(
    parent: any,
    {
      sessionsWhereInput,
      orderBy,
      skip,
      after,
      before,
      first,
      last,
    }: Selectors,
    { prisma }: Context
  ) {
    return prisma.sessions({
      where: sessionsWhereInput,
      orderBy,
      skip,
      after,
      before,
      first,
      last,
    });
  },
```
to
```ts
  sessions(
    parent: any,
    { where, orderBy, skip, after, before, first, last }: Selectors,
    { prisma }: Context
  ) {
    return prisma.sessions({
      where,
      orderBy,
      skip,
      after,
      before,
      first,
      last,
    });
  },
```

because in this example `sessionsWhereInput` is taken as a variable and not a type (it's in { }) and therefore, there was no `where` destructured variable passed down.

### DIFF
--- a/back/server/src/resolvers/query.ts
+++ b/back/server/src/resolvers/query.ts
@@ -6,20 +6,12 @@ import { Context, Selectors } from '../types';
 
 const Query = {
   blockNumbers(
-    parent: any,
-    {
-      blockNumberWhereInput,
-      orderBy,
-      skip,
-      after,
-      before,
-      first,
-      last,
-    }: Selectors,
+    _parent: any,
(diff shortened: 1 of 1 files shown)


yjkimjunior:
hmm it's been working fine for me, i.e. https://github.com/paritytech/Nomidot/pull/182

but i'll trust you

Tbaut:
Are you saying you're not having the problem?

PR #196, opened 2020-02-21
fix(context): Create a new {Api,System}Context on new provider

amaurymartiny:
Before: because we were using `useRef`, if we changed provider, it didn't create a new Api object (similar behavior with System).

After: we put api obect in state, and an effect to update state iif provider changed.


front/context/src/ApiContext.tsx
```diff
@@ -30,25 +27,32 @@ export function ApiContextProvider(
...
-    api.isReady.subscribe(() => {
+    const subscription = api.isReady.subscribe(() => {
       l.log(`Api ready, app is ready to use`);
 
-      setState({ isApiReady: true });
+      setIsReady(true);
     });
-  }, [api, provider]);
+
+    return subscription.unsubscribe();
```
yjkimjunior:
```suggestion
    return () => subscription.unsubscribe();
```

amaurymartiny: Ah yeah

front/context/src/ApiContext.tsx
```diff
@@ -30,25 +27,32 @@ export function ApiContextProvider(
...
+    const subscription = api.isReady.subscribe(() => {
       l.log(`Api ready, app is ready to use`);
 
-      setState({ isApiReady: true });
+      setIsReady(true);
     });
-  }, [api, provider]);
+
+    return () => subscription.unsubscribe();
+  });
```
yjkimjunior:
```suggestion
  }, [api]);
```
^ i would think this to be necessary

amaurymartiny:
Yep, you're right, fixed

Issue #188, opened 2020-02-19
front: Add to Cart


yjkimjunior:
User story:

* I want to select some validators to nominate without yet committing to anything
* I add them to a cart
* Before I commit (sign&submit the extrinsic), I want an overview of what I'm getting myself into (rewards wise)
* If I want to make edits, I can go back a page and do so.

PR #205, opened 2020-02-26
feat: add to cart, cart page

yjkimjunior:
implements #188

follow up PR required for adding store.js to automatically subscribe to localstorage.

fsz4RVHsmh

cart only shows 2 items because it won't add duplicates. There shouldn't be duplicates in the table in the first place though, see: https://github.com/paritytech/Nomidot/issues/206

Blocknumber is fetching because nodewatcher went haywire. restarted this morning, and it's flying through. should be synced by evening if it doesn't go tup again.


front/gatsby/src/apollo.ts
```diff
@@ -18,7 +18,7 @@ const wsLink = new WebSocketLink({
   options: {
     reconnect: true,
   },
-  webSocketImpl: process.env.NODE_ENV === 'production' ? ws : WebSocket, // ws does not work in the browser. Browser clients must use the native WebSocket object. But during production SSR, we must use ws.
+  webSocketImpl: typeof window === 'undefined' ? ws : window.WebSocket, // ws does not work in the browser. Browser clients must use the native WebSocket object. But during production SSR, we must use ws.
```
yjkimjunior: annoyingly this has very inconsistent behavior in dev mode...

yjkimjunior:
ok i'm just going to merge this and keep moving forrward if no one has anything to say....

front/gatsby/src/apollo.ts
amaurymartiny: could use IS_SSR

front/gatsby/src/components/Layout/Header/Header.tsx
```diff
@@ -192,54 +40,71 @@ export function Header(): React.ReactElement {
   }, [fetchAccounts]);
 
   useEffect(() => {
+    const count = getCartItemsCount();
+    setNumberOfItemsInCart(count);
     handleLogin();
+
+    // FIXME: use store.js for window.addEventListener('storage', updateCartItems);
```
amaurymartiny: "there's a hook for everything" https://usehooks.com/useLocalStorage/ (just fyi, not saying it's better than window.addEventListener)

amaurymartiny: i think you would need to put all cart items under one key, aka an array inside one locastroage key inside of `n` keys. But should be doable

PR #215, opened 2020-02-27
feat: localstorage listeners

yjkimjunior:


front/gatsby/src/util/helpers.ts
```diff
@@ -23,16 +24,14 @@ export function stripAddressFromCartItem(item: string): string {
   return item.slice(5);
 }
 
-export function getCartItemsCount(): number {
-  let count = 0;
+export function addToCart(stash: string, currCartCount: number): void {
+  writeStorage(`cart:${stash}`, stash);
```
amaurymartiny: Why not have one key in localStorage called `cart`, which is an array of stash addresses.
```
const [cart] = useLocalStorage('cart');
const cartCount = cart.length; // maybe you would need to JSON.parse(cart), i'm not sure
```

yjkimjunior: yeah, it is redundant, but I think read + parse would be more laggy than simply read + it's only a single key extra.

yjkimjunior:
ah shoot need to check IS_SSR for localstorage as well :(((

amaurymartiny:
Note: assume the user has 2 accounts, and wants to nominate different validators with each account: this data structure makes it difficult to switch accounts.

yjkimjunior:
> Note: assume the user has 2 accounts, and wants to nominate different validators with each account: this data structure makes it difficult to switch accounts.

maybe we should use IndexedDB instead then

amaurymartiny:
localStorage/IndexedDB are both very similar for this use case, localStorage could work too:
```
cart = {
  '{my_account1_stash}': ['validator_stash_1',...],
  '{my_account2_stash}': ['validator_stash_2',...]
}
```

The question is: long-term should we put the cart in AppDB? If yes, then localStorage/IndexedDB are temporary solutions, just how temporary are they? Would it be simpler to just do a React Context for now?

yjkimjunior:
>  localStorage could work too:
I think in localstorage that would look like:
```
cart = "{
  '{my_account1_stash}': "['validator_stash_1',...]",
  '{my_account2_stash}': "['validator_stash_2',...]"
}"
```
so `localStorage.setItem("cart", "{ my_account1_stash }: ['validator_stash_1']")` which hurts my eyes just looking at it. or maybe there's something i'm missing about localstorage.

> The question is: long-term should we put the cart in AppDB?
how so? Wouldn't something like Redis make more sense in that case? If so then I suppose we can simply use the address as a "userid"?

> Would it be simpler to just do a React Context for now?
I think it is currently as simple as it gets, but if we want to cater for the edge case where a user wants to add all validators at once and then select a subset to nominate from one account, then another subset from another, then yeah I guess React Context...but then we lose persistence...

yjkimjunior:
> or maybe there's something i'm missing about localstorage.

ok so as we discussed offline, @rehooks will handle serializing, so indeed that was the gap in my knowledge.

So todos left for this PR:

* [ ] - use this data structure ` '{my_account1_stash}': ['validator_stash_1',...],` with localstorage nothing more fancy required.
* [ ] - use `react-loadable` for SSR dynamic imports https://github.com/jamiebuilds/react-loadable

yjkimjunior:
TIL another hook https://github.com/alex-cory/use-ssr

yjkimjunior:
somehow production build works but development build can't parse typescript now

PR #227, opened 2020-03-07
added ingress controller with basic-auth

fevo1971:
Added an ingress controller to the k8s staging deployment. Mostly to give the *polkassembly* project access to the nodewatcher. It implements ssl and a http basic-auth.

### DIFF
--- a/front/kubernetes/staging/ingress.yaml
+++ b/front/kubernetes/staging/ingress.yaml
@@ -1,0 +1,29 @@
+apiVersion: v1
+data:
+  auth: YTokYXByMSRWdVhZTHVFSiRVMHZhZHJjUUw3YmJsZjdFRG1DUUIx
+kind: Secret
+metadata:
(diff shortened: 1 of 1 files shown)


front/kubernetes/staging/ingress.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v1
+data:
+  auth: YTokYXByMSRWdVhZTHVFSiRVMHZhZHJjUUw3YmJsZjdFRG1DUUIx
```
yjkimjunior: correct me if I'm wrong but we should probably not have secrets on Github?

fevo1971: it's already the encrypted password (+username) as a base64 encoded string. so it's kind of ok to check it in.

fevo1971:
this is no longer needed because the `nomidot-server` is now requested by polkassembly and it comes with it's own loadbalancer.

PR #229, opened 2020-03-08
Add treasury queries to server

Tbaut:
Nothing fancy. I've tested our Front with the server as a the chain-db graphql and this seems to be the only missing piece.


Tbaut:
merging as insubstantial.

Issue #233, opened 2020-03-09
Cannot read property 'at' of undefined


Tbaut:
```bash
$ node -r ts-node/register --max-old-space-size=8192 ./src/index.ts
Unknown types found, no types for ActiveEraInfo,Releases
2020-03-09 21:01:03    NODE-WATCHER: blockIndex: 1378323
2020-03-09 21:01:03    NODE-WATCHER: lastKnownBestFinalized: 1380087
2020-03-09 21:01:03    NODE-WATCHER: block: 1378323
2020-03-09 21:01:22    NODE-WATCHER: hash: 0x6c70cbdecaeb1a6b3507c8dcc9e437eb9d0f40a6d3f8a95fb1df66e993da5dfa
2020-03-09 21:01:40    NODE-WATCHER: bumped spec version to 1050, fetching new metadata
2020-03-09 21:02:53    NODE-WATCHER: Task --- createBlockNumber
2020-03-09 21:03:29 TASK: BLOCKNUMBER: NomidotBlock: {"authoredBy":"DxnzP6bkiF2SeQofK4o8shvgK5JyGsqrNjpd7qPtLZffiAG","hash":"0x6c70cbdecaeb1a6b3507c8dcc9e437eb9d0f40a6d3f8a95fb1df66e993da5dfa","startDateTime":1583776788000}
2020-03-09 21:03:29    NODE-WATCHER: Writing: {"authoredBy":"DxnzP6bkiF2SeQofK4o8shvgK5JyGsqrNjpd7qPtLZffiAG","hash":"0x6c70cbdecaeb1a6b3507c8dcc9e437eb9d0f40a6d3f8a95fb1df66e993da5dfa","startDateTime":1583776788000}
2020-03-09 21:03:29 TASK: BLOCKNUMBER: Prisma Block Number: {"number":1378323,"startDateTime":"2020-03-09T17:59:48.000Z","hash":"0x6c70cbdecaeb1a6b3507c8dcc9e437eb9d0f40a6d3f8a95fb1df66e993da5dfa","id":"ck7kydbbmw44x07191z8nat24","authoredBy":"DxnzP6bkiF2SeQofK4o8shvgK5JyGsqrNjpd7qPtLZffiAG"}
2020-03-09 21:03:29    NODE-WATCHER: Task --- createCouncil
2020-03-09 21:03:29    NODE-WATCHER: Writing: []
2020-03-09 21:03:29    NODE-WATCHER: Task --- createSession
[...]

PR #236, opened 2020-03-12
Make Nodewatcher run

Tbaut:
This is not nice, and should probably not be merged as is, but this is what I needed to do to get nodewatcher to run. It had many weird errors after the recent kusama upgrade.

It's a critical point in our infra.. and we're live so it needs to run.

Those were added manually because the pods were failing with "unknown module". Why suddenly.. I've no clue.
```js
    "@polkadot/api": "^1.6.0-beta.17",
    "@polkadot/types": "^1.6.0-beta.17",
    "@polkadot/util": "^2.6.1",
```

Those were commented out because the tasks were failing (#233 )
```bash
  // createStake,
  // createEra,
  // createNominationAndValidators,
```

Docker image `comment-out-v1.2.25` is running with this code.


Tbaut:
The comments are in master now :/

PR #240, opened 2020-03-17
fix(nodeWatcher): Handle promise errors

amaurymartiny:


back/node_watcher/src/nodeWatcher.ts
```diff
@@ -116,12 +116,8 @@ async function incrementor(
...
       const result = await task.read(blockHash, cached, api);
 
-      try {
-        l.warn(`Writing: ${JSON.stringify(result)}`);
-        await task.write(blockNumber, result);
-      } catch (e) {
-        l.error(e);
-      }
+      l.warn(`Writing: ${JSON.stringify(result)}`);
+      await task.write(blockNumber, result);
```
amaurymartiny: We were ignoring writing to Prisma errors before, hence skipping them, I'm not sure why. Now, if there's an error while writing to Prisma, we throw

back/node_watcher/src/nodeWatcher.ts
```diff
@@ -147,5 +143,5 @@ export async function nodeWatcher(tasks: NomidotTask[]): Promise<void> {
   const provider = new WsProvider(ARCHIVE_NODE_ENDPOINT);
   const api = await ApiPromise.create({ provider });
 
-  incrementor(api, provider, tasks);
+  return incrementor(api, provider, tasks);
```
amaurymartiny: incrementor is the culprit, it didn't have a `.catch`, and we got UnhandledPromiseRejectionWarning.

Now we return this, and it gets handled here: https://github.com/paritytech/Nomidot/blob/master/back/node_watcher/src/index.ts#L13

back/node_watcher/src/nodeWatcher.ts
Tbaut: I wonder if this will throw for errors such as unique constraints violated. I guess this is fine for the tip of the chain, but any resync with a modified schema would fail.
We can try still.

amaurymartiny: Ah right. Let's do small steps then, I reverted this.

PR #245, opened 2020-03-20
Remove yarn.lock from node_watcher and move it for dockerization

Tbaut:
- upgrade to 1.8.4 beta
- add @polkadot/api to package.json in node_watcher
- add a docker script to copy and remove yarn.lock form ../../ for dockerization.


yarn.lock
```diff
@@ -2471,6 +2487,24 @@
     eventemitter3 "^4.0.0"
     rxjs "^6.5.4"
 
+"@polkadot/api@1.8.0-beta.4", "@polkadot/api@^1.8.0-beta.4":
```
Tbaut: Isn't that a problem to have `1.8.0-beta.4` and `^1.8.0-beta.4`?
I've had this in the past and couldn't figure out whether it was ok or not..

amaurymartiny: No. The most important thing to have is one single `@polkadot/api` in yarn.lock

amaurymartiny: Which is actually not the case now

Tbaut:
ok, we now have docker scripts in `node_watcher` and `server`, local `yarn.lock` where removed. This is solution 1 from #245

PR #248, opened 2020-03-23
chore: bump deps

yjkimjunior:


back/node_watcher/src/tasks/createReferendum.ts
```diff
@@ -77,25 +81,31 @@ const createReferendum: Task<NomidotReferendum[]> = {
...
-          voteThreshold: referendumRawEvent.VoteThreshold,
-        };
-
-        l.log(`Nomidot Referendum: ${JSON.stringify(result)}`);
-        results.push(result);
+        if (isItOld) {
+
+        } else {
+          const result: NomidotReferendum = {
+            delay: (referendumInfo as ReferendumInfoTo239).delay,
```
Tbaut: ah damn I only see this now, this is fixed here: #251

Tbaut: Sorry I had to bump as well to get the correct types from @polkadot/api.

PR #257, opened 2020-03-25
Using our dedicated rpc-node

Tbaut:
it didn't work with `wss` or without the port number FYI.
Last-50K is running with it now.


Tbaut:
closing, it's in #261

PR #263, opened 2020-03-26
Fv dockerize in ci

fevo1971:
Added a new stage (dockerize) to the CI, that builds new docker Images

On **master** and **tags** for
- frontend 
- server
- noder_watecher

And uploads them to the Google Container Registry in the *test-installations* project.

Builds on **master** are tagged as `:master-<short-commit-hash>` and tagged branches
are tagged as: `<tag-name>-<short-commit-hash>`.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,8 +5,11 @@ stages:
   - dockerize
 
 variables:
-  GIT_DEPTH: '3'
-  SIMPLECOV: 'true'
+  GIT_DEPTH:      '3'
+  SIMPLECOV:      'true'
+  KUBE_NAMESPACE: "nomidot"
+  CI_REGISTRY:    "gcr.io/test-installations-222013"
+  DOCKER_TAG:     '$CI_COMMIT_REF_SLUG-$CI_COMMIT_SHORT_SHA'
 
 cache:
   key: '${CI_JOB_NAME}'
@@ -29,7 +32,7 @@ linux-test:
 
 #### stage:                        dockerize
 
-.docker_init:                   &docker_init
+.dockerize_init:                   &dockerize_init
   variables:
     DOCKER_DRIVER: overlay2
     DOCKER_HOST: tcp://localhost:2375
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -23,3 +29,57 @@ linux-test:
...
+  services:
+    - docker:dind
+  before_script:
+    - POD_NAME=$(echo ${CI_JOB_NAME} | sed -E 's/^[[:alnum:]:]+-//')
+    - export DOCKER_IMAGE="${CI_REGISTRY}/${KUBE_NAMESPACE}-${POD_NAME}"
+    - export DOCKER_IMAGE_FULL_NAME=$DOCKER_IMAGE:$DOCKER_TAG
+    - echo $DOCKER_IMAGE_FULL_NAME
+    - echo "$DOCKER_REGISTRY_JSON" | docker login -u _json_key --password-stdin https://gcr.io
+
+dockerize-frontend:
```
yjkimjunior: We've been using Netlify to deploy the frontend for the simplicity but with this available I think we can drop Netlify.

What do you think @amaurymartiny ?

amaurymartiny: Nice, cool! Yes, let's drop Netlify.

amaurymartiny: Although, the gitlab setup won't deploy a static frontend on each PR. Do we want to keep that?

fevo1971:
> thank you @fevo1971!
> 
> If I'm understanding this correctly, with this setup all our Kubernetes deployments need to do is have the ImagePullPolicy set and they'll update automatically whenever those relevant tags are merged to master?

It's a good question. My 1st approach here is to have a deployment to *staging* (aka dashboards-cluster-1) whenever `master` is updated. If you `tag` a release 
a deployment to *productions* (aka parity-prod) will be triggered. Docker images will be (docker)tagged either with `:master` or `:x.z.y` (for prod).

For staging a `imagePullPolicy: Always` will be set and for production to `IfNotPresent`.  

Will that work for you guys?

Later maybe only `docker build` in case of changes for a particular directory/file an master.
Like this:

```
dockerize-frontend:
  stage: dockerize
  environment:
    name: parity-build
  <<: *dockerize_init
  script:
    - docker build -t "$DOCKER_IMAGE_FULL_NAME" front
    - docker push "$DOCKER_IMAGE_FULL_NAME"
  rules:
    - if: '$CI_COMMIT_REF_SLUG == "master"'
      changes:
      - front/*
      - .gitlab-ci.yml
    - if: $CI_COMMIT_TAG
```

yjkimjunior:
Ok, that sounds good to me!

PR #264, opened 2020-03-29
cd before docker build

fevo1971:
Better change directory before executing `docker build`, it fails for *server* & *node_watcher* atm.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,7 +54,8 @@ dockerize-frontend:
     name: parity-build
   <<: *dockerize_init
   script:
-    - docker build -t "$DOCKER_IMAGE_FULL_NAME" front
+    - cd front
+    - docker build -t "$DOCKER_IMAGE_FULL_NAME" .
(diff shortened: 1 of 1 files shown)


yjkimjunior:
it looks like it fails to copy `yarn.lock` from the root now

fevo1971:
> it looks like it fails to copy `yarn.lock` from the root now

that was actually what i wanted to fix ;) but the reason for the copy to fail is obvious something else.
looks like there is no `back/node_watcher/yarn.lock` and no `back/server/yarn.lock` on master. right?

Tbaut:
No, so what we do when we dockerize manually is that we:
- cp ../../yarn.lock .
- docker build -t nomidot_watcher:v1.2.30 . && docker tag nomidot_watcher:v1.2.30 eu.gcr.io/test-installations-222013/nomidot_watcher:v1.2.30 && docker push eu.gcr.io/test-installations-222013/nomidot_watcher:v1.2.30
- rm yarn.lock

Issue #271, opened 2020-03-31
server: Investigate ECONNRESET


yjkimjunior:
frontend == ExternalIP ==> Loadbalancer ====> **Server Deployment ===ClusterIP ==> Prisma (Node Watcher) port 4466** problem area here

the error comes from the `server => prisma` networking.

so far we attempted:
1. turn it off and back on again (delete/replace server and nodewatcher deployments)
2. scale the server (add a replicaset)

neither seem to have actually solved the underlying issue.

trying this now....
3. scale the nodewatcher (double replicas)

yjkimjunior:
so it seems like the nomidotwatcher Loadbalancer Service is abruptly cutting the connection....

https://stackoverflow.com/questions/17245881/how-do-i-debug-error-econnreset-in-node-js#17637900

yjkimjunior:
ok so as we discussed on Riot https://github.com/kubernetes/kubernetes/issues/79365#issuecomment-572240923

it looks very likely that our issue is the nodewatcher is a pod made up of multiple containers, since we use the side-car model for GCP. 

according to that github comment, it means we need to explicitly set the resources we need.

otherwise, we get https://matrix.parity.io/_matrix/media/r0/download/matrix.parity.io/BlNsHgxHVFckekECbDLzOeYL

type of error.

Tbaut:
For the record, we used to have:
```bash
kubectl get hpa
NAME          REFERENCE                TARGETS                        MINPODS   MAXPODS   REPLICAS   AGE
nodewatcher   Deployment/nodewatcher   <unknown>/85%, <unknown>/80%   1         5         4          41d
```

Which I removed for now, to see if this happens again. We should configure an hpa correctly then.
![image](https://user-images.githubusercontent.com/33178835/78044490-0d50fa00-7375-11ea-81fb-0e042744f77b.png)

Tbaut:
Nodewatcher recently got a couple pods evicted, the GCP console says:
> Pod The node was low on resource: [MemoryPressure]. 

pod describe says the same:
> Message: The node was low on resource: memory. Container prisma was using 1278848Ki, which exceeds its request of 0. Container cloudsql-proxy was using 9364Ki, which exceeds its request of 0.

that's on last 50k. The running pod is as usual healthy and not showing any error. I'm worried about the "exceeds its request of 0." though

Tbaut:
It happened and it looks like there's a memory leak in primsa:
This is the nodewatcher pod
[...]

PR #281, opened 2020-04-01
feat: hpa and nodewatcher deployment resources

yjkimjunior:
https://github.com/paritytech/Nomidot/issues/271

https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/

### DIFF
--- a/back/node_watcher/kubernetes/nodewatcher-deployment-staging.yaml
+++ b/back/node_watcher/kubernetes/nodewatcher-deployment-staging.yaml
@@ -48,6 +48,13 @@ spec:
                 secretKeyRef:
                   name: cloudsql-db-credentials
                   key: dbname
+          resources:
+            limits:
+              cpu: 500m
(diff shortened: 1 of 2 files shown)


back/node_watcher/kubernetes/nodewatcher-deployment-staging.yaml
```diff
@@ -57,6 +64,13 @@ spec:
           volumeMounts:
             - name: cloudsql-instance-credentials
               mountPath: /secrets/cloudsql
+              resources:
+    limits:
+      cpu: 500m
+      memory: 1Gi
+    requests:
+      cpu: 50m
+      memory: 128Mi
```
Tbaut, line 73: indentation isn't right I guess.

yjkimjunior:
ok i'll whack the hpa since it's probably unncessary, but we could keep the resource specifications

PR #285, opened 2020-04-06
feat: cache accounts context results

yjkimjunior:
store the following in localstorage

* stashes
* derived balances
* derived staking info


yjkimjunior:
erm, not sure what happened to AddressSummary component with the new ui-components but it's  gotten....stretchy in a not so nice way. 

not something to åddress in this PR, but, note to self.

yjkimjunior:
also the caching removes some lag on initial render, but the information only loads as fast as AccountsContext is initilaized, which is only as fast as Api is initilaized. still, better than nothing in my experience.

front/context/src/AccountsContext.tsx
```diff
@@ -14,6 +14,7 @@ import { Option } from '@polkadot/types';
 import { AccountId, StakingLedger } from '@polkadot/types/interfaces';
 import { logger } from '@polkadot/util';
 import { decodeAddress, encodeAddress } from '@polkadot/util-crypto';
+import { writeStorage } from '@substrate/local-storage';
```
amaurymartiny, line 17: Oh, wasn't aware of this. Where is the source code?

front/context/src/AccountsContext.tsx
```diff
@@ -14,6 +14,7 @@ import { Option } from '@polkadot/types';
 import { AccountId, StakingLedger } from '@polkadot/types/interfaces';
```
amaurymartiny, line 14: fyi I'm using my own [AccountsContext](https://github.com/paritytech/substrate-light-ui/blob/master/packages/light-apps/src/components/ContextGate/context/AccountContext/AccountContext.tsx) in slui, so feel free to modify this one with breaking changes.

front/context/src/AccountsContext.tsx
yjkimjunior: https://github.com/yjkimjunior/local-storage

I just forked the rehooks one and made some small edits to a lingering PR to enable it for SSR

front/context/src/AccountsContext.tsx
yjkimjunior: ah ok, noted!

front/context/src/AccountsContext.tsx
amaurymartiny: Just a thought, not tested: if it's only to make `gatsby build` pass, maybe something like https://www.npmjs.com/package/localstorage-polyfill could be enough

PR #290, opened 2020-04-08
feat: extension signer

### LINKED ISSUE #212: front: submit bond
on `/accounts` page:

* [x] - click submit new bond
* [x] - choose some  account in extension to be the stash
* [x] - choose some other account in extension to be the controller
* [x] - sign & submit the extrinsic

yjkimjunior:
closes #212

* status notifier
* bug fixes and other improvements
* responsniveness


front/context/src/AccountsContext.tsx
```diff
@@ -178,6 +178,12 @@ export function AccountsContextProvider(props: Props): React.ReactElement {
     }
   }, [chain, originName]);
 
+  const setSigner = () => {
+    if (api && apiPromise && extension && isExtensionReady) {
+      api.setSigner(extension.signer);
+      apiPromise.setSigner(extension.signer);
```
amaurymartiny, line 184: Why are there 2 promise objects? More specifically, why is apiPromise needed?

yjkimjunior: it's not needed strictly, and it could be refactored if you think that's better. I was going over the tradeoffs of mixing both rx and promises and decided the readability and reduced boilerplate particularly with regard to one off queries with promises wins out over the "reactive everything" end of the stick.

I checked with apps of course to see if this happens there as well and indeed there is also a `withCall` as well as a `withObservable` mix depending on what's more convenient. From what I've read, the "reactive everything" is preferred because you can cancel subscriptions whereas you cannot with promises, but the way I'm using it so far is only for things like getting a nonce or calculating fees.

so tldr; `apiPromise` for `api.query...` and `api.consts` where sensible, and RX `api.tx`, to reduce boilerplate and increase readability (IMO).

amaurymartiny: > I checked with apps of course to see if this happens there as well and indeed there is also a withCall as well as a withObservable

apps has withCall and withObservable, but both of them use one ApiPromise under the hood.

In light-ui, I can see empirically that the API decoration takes a bit of time (the UI thread is blocked, it feels laggy for ~5s). With 2 APIs, there are 2 decorations. So in light-ui I'll go with 1 API object (the rx one). But I'm okay to create my own ApiContext for light-ui (as I did for AccountsContext).

PR #294, opened 2020-04-09
fix(context): Add local-storage dep

amaurymartiny:
- More correct dep graph: gatsby -> @context -> @local-storage
- add correct peer deps for @context


front/gatsby/package.json
```diff
@@ -10,8 +10,8 @@
     "@apollo/react-hooks": "^3.1.3",
     "@loadable/component": "^5.12.0",
     "@polkadot/extension-dapp": "^0.21.1",
+    "@substrate/context": "^0.3.23",
```
yjkimjunior, line 13: question :we have the webpack config that aliases this, so by adding it here I'm not sure which one "wins." Does it pull in the npm version or the local version. 

If it's the local version, why is this needed here?

If it's the npm version, IMO that makes development slightly more cumbersome, e.g. previously i changed a return type in txQueueContext, if gatsby had to pull the npm versionn, that little change would require a version bump and npm publish. If that's better practice then I can get used to it too.

front/context/package.json
```diff
@@ -17,7 +17,12 @@
     "start": "yarn build --watch",
     "test": "echo skipped"
   },
+  "dependencies": {
+    "@substrate/local-storage": "^2.2.9"
+  },
   "peerDependencies": {
+    "@polkadot/api": "^1.9.1",
+    "@polkadot/extension-inject": "^0.21.1",
```
yjkimjunior, line 25: if these are `peerDependencies`, then the gatsby app should install these as `dependencies` right?

front/gatsby/package.json
amaurymartiny: Since we added `@substrate/context` in the yarn workspace members, this never pulls from npm. To convince yourself, just have a look at your node_modules, there should be a little arrow in VSCode which means that the folder is just a symlink to the actual folder.

>  if gatsby had to pull the npm versionn

This should never happen because of yarn workspace.

The only reason we added the webpack config in gatsby is: by default, when you do `import { ... } from '@substrate/context'`, it looks into the `lib/` folder of `@substrate/context`'s folder (see package.json for where it's config). This works, but it also means that everytime you change something in `@substrate/context`, you need to build it again. This is what we do in light-ui.

By adding a webpack alias, the bundler interprets `from '@substrate/context'` with `from '../path/to/the/alias'`, which we pointed at the `src/*.ts` file. So when you have gatsby running, and change something in `@substrate/context`, it'll just hot-reload.

front/context/package.json
amaurymartiny: Yes, I made sure it's already the case.

front/gatsby/package.json
yjkimjunior: > Since we added @substrate/context in the yarn workspace members, this never pulls from npm.

ah, so that's what that arrow means. TIL, thanks for the clarification

I'm still not 100% w.r.t to this specific line though. If gatsby always symlinks to the `/path/to/alias` what exactly changes when you add this to `package.json`?

front/context/package.json
amaurymartiny: > is it enough to check it gets put into the root node_modules

Yes, node_modules is enough. Gatsby has `extension-dapp`, and `extension-dapp` automatically installs `extension-inject`

front/gatsby/package.json
amaurymartiny: > I'm still not 100% w.r.t to this specific line though. If gatsby always symlinks to the /path/to/alias what exactly changes when you add this to package.json?

Ah yeah, webpack probably would work too if we removed this line.

PR #295, opened 2020-04-10
fix: styling for tables and account page

yjkimjunior:
cleaning up to make it look presentable, responsive for mobile

idk wth is going on with addresssummary component... will copy it over here from Lichen in another PR so I have more control over it


yjkimjunior:
while details are visible in mobile, bonding action is hidden since you don't have the extension from there anyway. assumption is that mobile experience is read-only.

PR #298, opened 2020-04-11
feat: set cpu/mem requests and limits for prisma and cloudsql

yjkimjunior:
addresses #271

particularly the `exceeds its request of 0` as mentioned https://github.com/paritytech/Nomidot/issues/271#issuecomment-608330706

I had previously mistaken the mem/cpu requests fields to be only relate to HPA (which we agreed was unnecessary for our app), which was why that PR was closed.

This PR sets the requests for resources to be as following:

### prisma:
| | mem        | cpu        | 
| ------------- |:-------------:| -------------:| 
| request | 1.28GiB (failed at 1278848Ki === 1.219GiBi)      | 250m default (wasn't the cause of failure) |
| limit | 2.56GiB      | 500m |

### cloudsql
| |  mem        | cpu        | 
| ------------- |:-------------:| -------------:|
| request | 16Mi (failed at 9364Ki === 9.145Mi)     | 250m default (wasn't the cause of failure) |
| limit | 32Mi    | 500m  |

see also:

* https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
* https://medium.com/kubernetes-tutorials/efficient-node-out-of-resource-management-in-kubernetes-67f158da6e59
* https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/


yjkimjunior:
actually you did :) #271 (comment)

Tbaut:
Hmm, I think I didn't write this clearly. The pod describe was on the last 50k. I don't think the nodewatcher ever got evicted. But I can't remember exactly tbh.

yjkimjunior:
hmm but the containers that ran out of resources according to that log are in nodewatcher deployment.

PR #300, opened 2020-04-14
feat: nominate

yjkimjunior:
* [x] - nomination constants (nomination period, unbonding period)
* [x] - expected rewards (just a global average for now)
* [x] - fee, user input validations
* [x] - sign, submit extrinsic
* [x] - notification subscription

refactoring
* [x] - move constants,  to context
* [x] - reorg cart page components data flow
* [x] - components and typography (bring them in from ui-components once it gets there)


yjkimjunior:
I would update this with #302 once that's in.

PR #302, opened 2020-04-15
feat: Split ApiContext into ApiPromiseContext and ApiRxContext

Tbaut:
A couple questions:
- what the best way to test that
- do you run a script to publish and update the changelog or is that manual?
- I went for an object type, although you mentioned an array @amaurymartiny, lmk if you have a strong opinion on that.


front/context/src/ApiContext.tsx
```diff
@@ -19,43 +19,63 @@ export const ApiContext: React.Context<ApiContextType> = React.createContext(
...
 }
 
 export function ApiContextProvider(
   props: ApiContextProviderProps
 ): React.ReactElement {
-  const { children = null, provider, promise } = props;
+  const {
+    children = null,
+    provider,
+    type = { withRxjs: true, withPromise: false },
```
Tbaut, line 39: I went for a default `withRxjs=true` to keep backward compatibility

front/context/src/ApiContext.tsx
```diff
@@ -19,43 +19,63 @@ export const ApiContext: React.Context<ApiContextType> = React.createContext(
   {} as ApiContextType
 );
 
+export interface ApiType {
+  withPromise?: boolean;
+  withRxjs?: boolean;
```
amaurymartiny, line 24: Since the Api is called ApiRx, i would call it `withRx` simply

front/context/src/ApiContext.tsx
```diff
@@ -19,43 +19,63 @@ export const ApiContext: React.Context<ApiContextType> = React.createContext(
...
 ): React.ReactElement {
-  const { children = null, provider, promise } = props;
+  const {
+    children = null,
+    provider,
+    type = { withRxjs: true, withPromise: false },
+  } = props;
   const [api, setApi] = useState<ApiRx>(new ApiRx({ provider }));
   const [apiPromise, setApiPromise] = useState<ApiPromise>(
     new ApiPromise({ provider })
```
amaurymartiny, line 43: This is where the decoration happens (can be a bit heavy on UI), so we need to `new` them only if the corresponding `with` is set

amaurymartiny:
> what the best way to test that

I would just run `yarn start` and see if everything works.

> do you run a script to publish and update the changelog or is that manual?

Yes, from the root folder, `lerna version --conventional-commits && yarn lerna:publish`, it does everything
    
> I went for an object type, although you mentioned an array @amaurymartiny, lmk if you have a strong opinion on that.

the with one is better 👍

Tbaut:
I'll give a try to the following, although a little more convoluted than the 2 context approach.
> we coerce some TS typings into useApiRx and useApiPromise

Tbaut:
As I'm about to code it, I realize that the props `withApiPromise` and `withApiRx` aren't accessible to users, this is set here in the context provider, and can't be passed as argument in a hook.

Unless I miss something, you can't do this, because you don't know what's asked for.
>export useApiRx = useContext(ApiContext);
export useApiPromise = useContext(ApiContext);

But if we let `ApiContext` create and return both the `apiPromise` and the `apiRx`, then custom hooks `useApiPromise` and `useApiRx` can filter to return one or the other, like
```js
export function useApiRx() {
  const { apiRx } = useContext(ApiContext);
  return { api: apiRx };
}

export function useApiPromise() {
  const { apiPromise } = useContext(ApiContext);
  return { api: apiPromise };
}
```

It feels super wasteful (or is it simply super bad?) because `ApiContext` will always create both apis.

Unless I miss something, I guess we should go for 2 contexts (and we can have 2 hooks to be fancy).

amaurymartiny:
> As I'm about to code it, I realize that the props withApiPromise and withApiRx aren't accessible to users, this is set here in the context provider, and can't be passed as argument in a hook.

The context provider is also available to the user, he actually needs to add it himself:
```ts
import { ApiContextProvider } from '@substrate/context';

<ApiContextProvider withRx >
  // somewhere in a component
  const { api } = useApiRx();
</ApiContextProvider
```

> Unless I miss something, I guess we should go for 2 contexts (and we can have 2 hooks to be fancy).

While the above with one context could work (with TS typings), I actually also now think 2 contexts would be the most straightforward way

Tbaut:
Thanks for the clarification, I was confused with the provider being needed. I'll do 2 providers and 2 hooks then.

Tbaut:
We're using the  (old name) `ApiContext` in `AccountContext`. 
You get the apiRx (named `api`) [here](https://github.com/paritytech/Nomidot/blob/master/front/context/src/AccountsContext.tsx#L65) and only use it once to set the Signer [here](https://github.com/paritytech/Nomidot/blob/master/front/context/src/AccountsContext.tsx#L182-L183).

Since `api` aka `apiRx` is never used elsewhere, I don't think it's actually needed is it? In the case, in my PR I'd just use `ApiPromiseContext` instead and remove the signer par with `apiRx`.

Did I miss something ?

Tbaut:
Additional question for the hooks. I guess these don't belong to `/context`. I'd leave them for now. We can create a `@substrate/react-hooks` later on.

Tbaut:
I'll use it right away in Polkassembly as well and get rid of our implementation.

Tbaut:
I've used @substrate/context now, `yarn start` itself doesn't give any error, but visiting `localhost:8000` throws the react error:
> Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Is that my changes or something else @yjkimjunior ?

yjkimjunior:
well everywhere in the app that uses the context will have to be adapted to your changes.

I can do that and push to your branch if you want

Tbaut:
ah yeah, I expected a `yarn start` error I guess. doing the change now, I didn't think this far.

yjkimjunior:
if you run yarn start in root it would error :)

Tbaut:
I did run it form the root, it didn't error out. I didn't expect to have the context used anywhere else than in the wrapper. The 2 other context I found didn't get any change from this PR (TxQueue and system).

Also there's a 
```json
"paths": {
      "@substrate/context": ["front/context/src"]
    }
```

That should probably not stay (unless for development).

In any case, this error seems not really related.. but I can't really tell. I'd be glad if you take a look

yjkimjunior:
I get the same error in local as in the CI, i.e. ApiContext is not exported from @substrate/context, because it indeed isn't. So the gatsby app needs to be updated to use the updated context as you did for this PR

Tbaut:
Thanks a lot, this is getting big! I can't review my own PR, so feel free to merge if you're confident.

Tbaut:
Feel free to merge when you're happy with it @yjkimjunior 

>Yes, after this PR gets merged, once you think it's worth a publish: run from the root folder, master branch, lerna version patch --conventional-commits && yarn lerna:publish, it does everything

BTW I guess this is worth a `lerna version major` since it breaks compatibility.

yjkimjunior:
> Feel free to merge when you're happy with it @yjkimjunior
> 
> > Yes, after this PR gets merged, once you think it's worth a publish: run from the root folder, master branch, lerna version patch --conventional-commits && yarn lerna:publish, it does everything
> 
> BTW I guess this is worth a `lerna version major` since it breaks compatibility.

published! https://github.com/paritytech/Nomidot/pull/307/

Tbaut:
Thanks, I just checked and the latest is still 0.3.24 https://www.npmjs.com/package/@substrate/context

PR #307, opened 2020-04-20
chore: bump v0.4.0

yjkimjunior:


amaurymartiny:
I would advocate to use `yarn lerna version minor --conventional-commits` or `yarn lerna version 0.4.0 --conventional-commits`, so that it gets bumped everywhere

PR #309, opened 2020-04-20
Fix double api instanciation in @substrate/context

Tbaut:
Prevent the api to be instanciated twice.
This changes quite some things in Nomidot, feel free to add anything to this branch @yjkimjunior as you see fit.

### DIFF
--- a/front/context/src/ApiRxContext.tsx
+++ b/front/context/src/ApiRxContext.tsx
@@ -9,7 +9,7 @@ import React, { useEffect, useState } from 'react';
 import { ApiRxContextProviderProps } from './types';
 
 export interface ApiRxContextType {
-  api: ApiRx; // From @polkadot/api\
+  api: ApiRx | undefined; // From @polkadot/api\
   isApiReady: boolean;
 }
 
@@ -23,7 +23,7 @@ export function ApiRxContextProvider(
   props: ApiRxContextProviderProps
 ): React.ReactElement {
   const { children = null, provider } = props;
-  const [apiRx, setApiRx] = useState<ApiRx>(new ApiRx({ provider }));
+  const [apiRx, setApiRx] = useState<ApiRx>();
   const [isReady, setIsReady] = useState(false);
 
   useEffect(() => {
@@ -38,14 +38,18 @@ export function ApiRxContextProvider(
     // We want to fetch all the information again each time we reconnect. We
     // might be connecting to a different node, or the node might have changed
     // settings.
-    const subscription = apiRx.isReady.subscribe(() => {
-      l.log(`Api ready.`);
+    let subscription: any;
 
-      setIsReady(true);
-    });
+    if (apiRx) {
+      subscription = apiRx.isReady.subscribe(() => {
+        l.log(`Api ready.`);
 
-    return (): void => subscription.unsubscribe();
-  }, [apiRx.isReady]);
+        setIsReady(true);
+      });
+    }
+
+    return (): void => subscription && subscription.unsubscribe();
+  }, [apiRx]);
 
   return (
     <ApiRxContext.Provider value={{ api: apiRx, isApiReady: isReady }}>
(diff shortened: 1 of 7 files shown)


front/context/src/ApiRxContext.tsx
```diff
@@ -38,14 +38,18 @@ export function ApiRxContextProvider(
     // We want to fetch all the information again each time we reconnect. We
     // might be connecting to a different node, or the node might have changed
     // settings.
-    const subscription = apiRx.isReady.subscribe(() => {
-      l.log(`Api ready.`);
+    let subscription: any;
```
Tbaut, line 41: This is not nice, neither I nor the quick fix could find the type `Subscription`...

front/gatsby/src/components/Cart/CartItems.tsx
```diff
@@ -59,6 +59,11 @@ const CartItems = (): React.ReactElement => {
     }
   };
 
+  if (!api || !isApiReady) {
```
Tbaut, line 62: This is not perfect either, you might want to change this @yjkimjunior

front/gatsby/src/components/Layout/Header/Subheaders/StakingHeader.tsx
```diff
@@ -14,7 +13,6 @@ import { StakingHead } from '../types';
 const StakingHeader = (): React.ReactElement => {
   const { data } = useSubscription(STAKING_SUBSCRIPTION);
   const [stakeHead, setStakeHead] = useState<StakingHead>();
-  const { api } = useContext(ApiRxContext);
```
Tbaut, line 17: I figured out the api was never used in this component.

front/gatsby/src/pages/validators.tsx
```diff
@@ -112,6 +112,11 @@ const CurrentElectedList = (): React.ReactElement => {
   };
 
   const renderValidatorsTable = (): React.ReactElement => {
+    if (!api || !isApiReady) {
+      console.warn('Api not set');
+      return <></>;
```
Tbaut, line 117: Same as above, this is not ideal, you may want to change this.

front/context/src/ApiRxContext.tsx
```diff
@@ -9,7 +9,7 @@ import React, { useEffect, useState } from 'react';
 import { ApiRxContextProviderProps } from './types';
 
 export interface ApiRxContextType {
-  api: ApiRx; // From @polkadot/api\
+  api: ApiRx | undefined; // From @polkadot/api\
```
amaurymartiny, line 12: Ah, so I wanted to avoid exactly this, in my Zulip comments, since the `api &&` everywhere in the code is kind of useless.

But okay to play along if @yj's okay.

Tbaut: I've a hard time building a mental model for that. When you use the `{ api } = useContext(...)`, the `new ApiPromise(..)` (or Rx) is a blocking call and will only return with an `ApiPromise` instance, so that your `api` is never undefined, but you have to wait?

On the other hand, I don't see how we can go around this nicely. if we do:
```js
  const [apiPromise, setApiPromise] = useState<ApiPromise>();
  const [isReady, setIsReady] = useState(false);

  useEffect(() => {
    if (!apiPromise) {
      setApiPromise(new ApiPromise({ provider }));
      setIsReady(false);
    }
  }, [provider]);
```

to prevent the first `useEffect`, I wonder if what will happen when you change provider. This is something you should test but the `apiPromise` might be defined, and there will be no new instantiation.

Tbaut:
Closing, I'd love to get an explanation regarding https://github.com/paritytech/Nomidot/pull/309#discussion_r411501163 as I don't think I get it properly.

front/context/src/ApiRxContext.tsx
amaurymartiny: The `new ApiPromise(..)` blocking call is actually not done at the step `{ api } = useContext(...)`, but at the step `<ApiPromiseContextProvider />`, i.e. when you mount the provider component.

Once this component is mounted, and the blocking call is finished, then api will always be defined, so `{ api } = useContext(...)` (instant call, no blocking code) should give a defined api.

Tbaut: understood thanks, the `new ApiPromise(..)` is the blocking call, and it's called once as long as the provider isn't changed.

PR #311, opened 2020-04-21
fix(context): Don't instantiate API twice

amaurymartiny:
superseded #309


amaurymartiny:
Merging, will try on light-ui asap

Issue #310, opened 2020-04-21
refactor: Contexts


yjkimjunior:
1. there's a bit of an antipattern going on in our context package using hooks and context, preventing appropriate rerenders on context state mutation.

https://medium.com/simply/state-management-with-react-hooks-and-context-api-at-10-lines-of-code-baf6be8302c

2. functions used by `useEffect` hooks should be moved into the hook themselves, and wrapped in a `useCallback` or `useMemo` as appropriate.

PR #312, opened 2020-04-21
feat: refactor accounts context 

yjkimjunior:
addressing first part of #310

so this indeed fixes the issue with changes in context not propagating down via callbacks which is good.

- [x] : remove `useState`, replace with `useReducer`
- [x] : change `set...State` to `dispatch: {type: ..., value: ...}`
- [x] : simply pass the `state` and `dispatch` through `Provider`
- [x] : make sure it doesn't break SSR
- [x] : make appropriate changes down the component tree


CHANGELOG.md
```diff
@@ -3,6 +3,26 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.4.2](https://github.com/paritytech/Nomidot/compare/v0.4.1...v0.4.2) (2020-04-22)
```
amaurymartiny, line 6: It's really better to run `learn version *` from master branch. Right now this CHANGELOG shows all your internal commits, which will disappear once we squash this PR, and the v0.4.2 tag won't point to the correct commit.

amaurymartiny: If this workflow is confusing, I'm happy to revisit.

yjkimjunior: ah shoot sorry... I thought that `lerna version` just bumps the version in package.json and Changelog and then we're only to do `lerna publish` from master.

Should we put this in CI?

yjkimjunior: should I revert that commit?

amaurymartiny: > Should we put this in CI?

Yeah, that's what I was thinking: let CI do `yarn lerna version && yarn publish`. Maybe we could put a label `release:{minor,major,patch}`? I'm pretty sure Github Actions can do that easily

amaurymartiny: > should I revert that commit?

That would be the cleanest for future people reading the changelog:
- revert that commit
- delete the 0.4.2 tag from github
- run the whole process again from master once this is merged

amaurymartiny: Alternatively, for the CI, you could what Jaco does: `yarn lerna:version` (without --conventional-commits) manually like in this PR, manually update CHANGELOG, and let CI do the publishing. The scripts for that are already available.

yjkimjunior: > The scripts for that are already available.

I personally prefer your first suggestion, but also what extra scripts are needed for this?

amaurymartiny: If we were to follow exactly Jaco's way, I believe none, they are all here https://github.com/polkadot-js/dev/tree/master/packages/dev/scripts

yjkimjunior: > > should I revert that commit?
> 
> That would be the cleanest for future people reading the changelog:
> 
> * revert that commit
> * delete the 0.4.2 tag from github
> * run the whole process again from master once this is merged

I've so far reverted the commit and deleted the tag from github.

Next I'll get https://github.com/paritytech/Nomidot/pull/314 through and then test by merging this branch

yjkimjunior: if we do add `release:{minor,major,patch}` on PRs to determine the version bump, then we'd no longer use the `--conventional-commits` flag, right (since `lerna` autodetects the bumps based on fix/feat/breaking in commit messages)?

I'm fine with either, but the auto-generated Changelog is quite helpful.

yjkimjunior: https://github.com/paritytech/Nomidot/pull/314

amaurymartiny: You can let lerna autodetect the bump, that's `lerna version --conventional-commits`.

But you can also force lerna to bump as you decide:
- `lerna version patch --conventional-commits` bumps a semver patch (even though some commit messages are feat or have breaking change) and adds an entry in the changelog

So my proposal would be:
- never do `lerna version --conventional-commits`
- always do `lerna version {major,minor,patch} --conventional-commits` based on the PR label `release:{major,minor,patch}`

yjkimjunior: do you think it'd be better to do this via Github Actions, or via Gitlab? I just went with gitlab since that's where all the CI stuff is already.

amaurymartiny: I personally tend to prefer github actions, for 3 reasons: it's faster than gitlab, there's a bigger community (with their marketplace), and I generally can figure things myself (whereas I generally need devops team help with Gitlab).

But both should be able to do the work fine.

Also: if you go with Github, you would need to put your NPM_TOKEN on Github, in their Secrets. I think it's fine (Jaco does it, I did with js-libs before), but it's of course better to store this secret on our own Gitlab.

Issue #313, opened 2020-04-22
CI: lerna version and publish on release:{minor,major,patch} PRs


yjkimjunior:
put config to do lerna version yarn `lerna:version {patch, minor, major} && yarn publish` on PRs merged to master with a label `release:{minor,major,patch}`

see: https://github.com/paritytech/Nomidot/pull/312#discussion_r412879514

PR #314, opened 2020-04-22
fix(#313): gitlab for lerna bump and publish

yjkimjunior:
addressing #313 

add a stage in CI to run `lerna version bump --conventional-commits && lerna publish`  on merge to master of any PR on a branch name ending with `-release`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -34,32 +34,12 @@ linux-test:
 lerna-patch:
   stage: release
   script:
-    - yarn lerna:bump patch
+    - yarn lerna:bump
     - yarn lerna:publish 
   only:
     - master
     - tags
-    # - /^-release:patch.*$/
-
-lerna-minor:
-  stage: release
-  script:
-    - yarn lerna:bump minor
-    - yarn lerna:publish 
-  only:
-    - master
-    - tags
-    # - /^-release:minor.*$/
-
-lerna-major:
-  stage: release
-  script:
-    - yarn lerna:bump major
-    - yarn lerna:publish 
-  only:
-    - master
-    - tags
-    # - /^-release:major.*$/
+    # - ^\w.*-release.*$
 
 #### stage:                        dockerize
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,8 +38,7 @@ lerna-patch:
     - yarn lerna:publish 
   only:
     - master
-    - tags
-    # - ^\w.*-release-patch.*$
+    # - /^\w.*-release-patch.*$/
 
 lerna-minor:
   stage: release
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -30,6 +31,36 @@ linux-test:
   tags:
     - linux-docker
 
+lerna-patch:
+  stage: release
+  script:
+    - yarn lerna:bump patch
```
amaurymartiny, line 37: With `--conventional-commits` to add an entry to the changelog?

.gitlab-ci.yml
```diff
@@ -30,6 +31,33 @@ linux-test:
...
+    - yarn lerna:publish 
+  only:
+    - master
+    # - /^\w.*-release-patch.*$/
+
+lerna-minor:
+  stage: release
+  script:
+    - yarn lerna:bump minor
```
amaurymartiny, line 46: Do we need a yarn install here? Or does this run after a previous job that already did yarn install?

.gitlab-ci.yml
```diff
@@ -30,6 +31,33 @@ linux-test:
...
     - linux-docker
 
+lerna-patch:
+  stage: release
+  script:
+    - yarn lerna:bump patch
+    - yarn lerna:publish 
+  only:
+    - master
+    # - /^\w.*-release-patch.*$/
```
amaurymartiny, line 41: Should this be uncommented?

yjkimjunior:
i'll close this PR now actually since we agreed to go with Github Actions just now

.gitlab-ci.yml
yjkimjunior: fwiu the stages run in order as they are defined, i.e. test, release, dockerize

.gitlab-ci.yml
yjkimjunior: if we have `--conventional-commits` then we cannot specify `{patch, minor, major}` (it'll be inferred by lerna).

yjkimjunior: so actually I would vote to have a single gh-action which uses `--conventional-commits` so that we can autogenerate the CHANGELOG

amaurymartiny: > if we have --conventional-commits then we cannot specify {patch, minor, major} (it'll be inferred by lerna).

We still can! See the part "you can also force lerna to bump as you decide" https://github.com/paritytech/Nomidot/pull/312#discussion_r413631345

amaurymartiny: That's personally what I've been doing in the 0.3.* releases on Nomidot. A lot of PRs had `feat:`, but I did `lerna version patch --conventional-commits` to stay on 0.3.*.

`--conventional-commits` should be understood as:
- if there's no `{path,minor,major}`, then I'll infer the bump based on commit based, and populate the changelog
- if you specify one manually, then I'll just populate the changelog

yjkimjunior: ah, neat! that was not at all clear from the docs 😅  https://github.com/lerna/lerna/tree/master/commands/version#--conventional-commits

PR #315, opened 2020-04-23
github action: release.yml

yjkimjunior:
bump and publish lerna


yjkimjunior:
seems to be passing a dry run with https://github.com/nektos/act, will test with #316

PR #320, opened 2020-04-27
feat: v1.5 sync config

yjkimjunior:
not included in this PR is the updated prisma config secret.

viewable by: `kubectl get secrets prisma-config-v1.5 -o yaml -n nodewatcher-staging`
decode with: `"echo {output from above} | base64 --decode"`

```
port: 4466
databases:
  v1-5:
    connector: postgres
    host: 127.0.0.1
    user: DB_ADMIN
    password: DB_PASSWORD
    rawAccess: true
    port: 5432
    migrations: true
    connectionLimit: 8
```

this is almost exactly the same as with v1 that's been syncing. importantly `default` was changed to v1-5 matching the `dbname` in our gcloud sql project. https://console.cloud.google.com/sql/instances/kusama-etl/databases?project=test-installations-222013

while this uses the same port 4466, the `nodewatcher-v1.5` deployment gets assigned its own ClusterIP so it shouldn't matter.

### DIFF
--- a/back/node_watcher/kubernetes/v1.5/nomidotwatcher-job.yaml
+++ b/back/node_watcher/kubernetes/v1.5/nomidotwatcher-job.yaml
@@ -27,7 +27,7 @@ spec:
       restartPolicy: Never
       containers:
         - name: nomidotwatcher
-          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:23dfc5a9638f00e5b3cac05abe46b59bf7f69f108f2974eb880a88e89bab66bb
+          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:f88f02e79cfc2c5a8f59ab2fae50f26eda40f1bd7ee4cc8c8b07b194c11d2cee
           imagePullPolicy: Always
           env:
             - name: PRISMA_ENDPOINT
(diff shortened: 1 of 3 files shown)


back/node_watcher/kubernetes/v1.5/nomidotwatcher-job.yaml
```diff
@@ -27,7 +27,7 @@ spec:
       restartPolicy: Never
       containers:
         - name: nomidotwatcher
-          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:23dfc5a9638f00e5b3cac05abe46b59bf7f69f108f2974eb880a88e89bab66bb
+          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:f88f02e79cfc2c5a8f59ab2fae50f26eda40f1bd7ee4cc8c8b07b194c11d2cee
```
yjkimjunior, line 30:
```suggestion
          image: eu.gcr.io/test-installations-222013/nomidot_watcher@sha256:086488aee19b982f83c557d6835c02fd1324affe9fb8b77b91b9f692413ddd7b
```

yjkimjunior:
ditching this in favor of polkascan

PR #321, opened 2020-04-27
fix: memory leak(s)

yjkimjunior:
- [x] : in main loop, use `let` instead of `const`, set reference to null when out of scope
- [ ] : in each task, clean up any intermediate references
- [ ] : make sure to remove event listeners once they go out of scope if there are any
- [ ] : make sure only just as many network connections are opened as needed

see also:
- https://medium.com/@_lrlna/garbage-collection-in-v8-an-illustrated-guide-d24a952ee3b8
- https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27#.pmqejrn8q
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management
- https://nodejs.org/en/docs/guides/debugging-getting-started/
- https://www.ibm.com/developerworks/web/library/wa-memleak/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # K8s secrets
+**/**/v1.5/config.yaml
 **/**/cloudsql-secrets.yaml
 **/**/prisma-secrets.yaml
 
--- a/back/node_watcher/src/nodeWatcher.ts
+++ b/back/node_watcher/src/nodeWatcher.ts
@@ -4,12 +4,13 @@
 
 import { ApiPromise, WsProvider } from '@polkadot/api';
 import { getSpecTypes } from '@polkadot/types-known';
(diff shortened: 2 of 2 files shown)


Tbaut:
Did you get some noticeable improvements with this already?

yjkimjunior:
it's 4-5k fewer references but that could be attributable to non-determinsm in the garbage collection. will test again after i finish the second checkpoint.

yjkimjunior:
it looks...slightly better. but not enough.

Retained Size:
**Before**
Object: 46%
Array: 43%
closures: 36%

**After**
Object: 37%
Array: 38%
closures: 34%

yjkimjunior:
not going to waste more time on this approach for now, will reassess after https://github.com/paritytech/Nomidot/pull/324 which is probably the better solution.

PR #322, opened 2020-04-27
Bump api to 1.12.0@beta-10

Tbaut:


yarn.lock
```diff
@@ -2293,40 +2441,75 @@
...
+    "@polkadot/types" "1.11.2"
+    "@polkadot/types-known" "1.11.2"
+    "@polkadot/util" "^2.8.1"
+    "@polkadot/util-crypto" "^2.8.1"
     bn.js "^5.1.1"
     eventemitter3 "^4.0.0"
-    rxjs "^6.5.4"
+    rxjs "^6.5.5"
+
+"@polkadot/api@1.12.0-beta.10", "@polkadot/api@^1.12.0-beta.10":
```
amaurymartiny, line 2495: There are two versions of api in node_modules, would be good to use 1.12-beta everywhere

Tbaut: Should be better now, there's only `@polkadot/api@1.12.0-beta.10`

Issue #267, opened 2020-03-30
Investigate memory leak in node watcher


Tbaut:
![image](https://user-images.githubusercontent.com/33178835/77887134-5700ee00-726a-11ea-80d7-727acf0b66ac.png)

a pod describe shows:
```bash
Status:         Failed
Reason:         Evicted
Message:        The node was low on resource: memory. Container nomidotwatcher was using 763096Ki, which exceeds its request of 0. 
```

PR #324, opened 2020-04-27
Move node watcher task to cloud function

niklabh:
Possible solution for memory leak: https://github.com/paritytech/Nomidot/issues/267

We can move running task read write part to cloud function (https://cloud.google.com/functions). Cloud function are serverless compute instances which starts and tears down for each invokation. So memory leak issue will not arise. 

Long running script will just invoke cloud function with blockIndex.

To test on local:

Start task runner/cloud function:
```
ARCHIVE_NODE_ENDPOINT=ws://127.0.0.1:9944 PRISMA_ENDPOINT=http://0.0.0.0:4466 PORT=4489 yarn start:runner
```

Start long running loop:
```
MAX_LAG=5 BLOCK_IDENTIFIER=aasdsdasdadssa ARCHIVE_NODE_ENDPOINT=ws://127.0.0.1:9944 TASK_RUNNER_SERVER=http://localhost:4489 PRISMA_ENDPOINT=http://0.0.0.0:4466 yarn start
```

long running loop will invoke cloud function with incrementing blockIndex. Cloud function will execute and tear down releasing memory. Invokation rate will be at the speed of block production.

### DIFF
--- a/back/node_watcher/docker-compose.yml
+++ b/back/node_watcher/docker-compose.yml
@@ -2,18 +2,19 @@ version: '3.7'
 services:
   main:
     build: ./
-  postgres:
-    image: postgres:12.1
-    restart: always
-    ports:
-      - '5432:5432'
-    environment:
-      POSTGRES_USER: prisma
-      POSTGRES_PASSWORD: prisma
-    volumes:
-      - postgres:/var/lib/postgresql/data
+  # postgres:
+  #   image: postgres:12.1
+  #   restart: always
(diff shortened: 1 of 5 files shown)


niklabh:
Need help with setting up cloud functions. I need permissions

niklabh:
It will not slow down task runner. Cloud function are desinged in such a way that they take time only for some initial invokations after that cloud functions are considered warmed up and next invokations takes only some milliseconds.

PR #326, opened 2020-04-29
fix: dont try to write on unique constraint invariant

yjkimjunior:
instead of relying on prisma to error and not write, just check existence before write...

it should get rid of those `Error: Unique Constaint Invariant Violation`s


yjkimjunior:
> I'd be for: adding a monitoring (just to make sure nodewatcher is running) and removing all the logs to actually log only the errors.

I'd be +1 for that as well, but would instead make the logs more compact, like just the task name and the unique id.

PR #342, opened 2020-05-05
feat: beef up validators table data

yjkimjunior:
loads in about 3 seconds initially, almost instantly with the cache

Screenshot 2020-05-06 at 18 48 42

follow up PRs to come for the header area tooltips, sortable columns


front/gatsby/src/apollo.ts
```diff
@@ -9,7 +9,7 @@ import { WebSocketLink } from 'apollo-link-ws';
 import { getMainDefinition } from 'apollo-utilities';
 import * as ws from 'ws';
 
-// TODO: 
+// TODO:
```
yjkimjunior, line 12:
```suggestion
// TODO: get a production env and domain, i.e. decide a name and buy a domain
```

PR #343, opened 2020-05-05
fix: change endpoint in dev env

yjkimjunior:
change frontend w.r.t. https://github.com/paritytech/Nomidot/pull/340

### DIFF
--- a/front/gatsby/src/apollo.ts
+++ b/front/gatsby/src/apollo.ts
@@ -21,7 +21,7 @@ const wsLink = new WebSocketLink({
   uri: `${
     process.env.NODE_ENV === 'production'
       ? 'ws://nomidot-server.nomidot-staging:4000'
-      : 'ws://test.nodewatcher-server.polkassembly.io/'
+      : 'wss://test.nodewatcher-server.polkassembly.io/'
   }`,
   options: {
     reconnect: true,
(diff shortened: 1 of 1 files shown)


front/gatsby/src/apollo.ts
```diff
@@ -13,15 +13,15 @@ const httpLink = new HttpLink({
   uri: `${
     process.env.NODE_ENV === 'production'
       ? 'http://nomidot-server.nomidot-staging:4000'
```
Tbaut, line 15: I don't think this will work in prod. This is a local url that only works in staging.

niklabh: This works in k8s. Internal urls are of format service-name.namespace

Tbaut: But `prod` and `staging` clusters are not connected, in the prod cluster, you won't be able to access `http://...**nomidot-staging**` that lives in the other cluster.

yjkimjunior: ah indeed, so I'd change `process.env.NODE_ENV === 'staging'` to use `'http://nomidot-server.nomidot-staging:4000'`, then use another one for `production` once we get there.

Is that how you do it in Polkassembly @Tbaut?

Tbaut: What we have in Polkassembly is `https://test.nodewatcher-server.polkassembly.io/` for prod and the local `http://nomidot-server.nomidot-staging:4000` for test. Although the `test.nodewatcher-server.polkassembly.io` is available from anywhere, so you could use that both in prod and test, we've tested it successfully yesterday.

PR #349, opened 2020-05-07
chore: bump @polkadot/api to 1.13.1

Tbaut:
Btw we're now using `@substrate/context` in prod in Polkassembly :)
- ran `yarn start` on Gatsby
- ran the node watcher

Can you please release a new @substrate/context version then?


Tbaut:
Absolutely, although if I'm not mistaking, the fact that it's asking for a minimum version actually saved us today as our front had an older version.

PR #353, opened 2020-05-12
fix(context): Reset syncing to true when chaning provider

amaurymartiny:
Subtle bug: when in light-ui I'm changing providers: if the old provider was syncing or not syncing, it should not have any effect on the syncing status of the new provider


amaurymartiny:
btw, should I have renamed my branch to something special to make it publish after merge?

I'm okay to do it manually too

yjkimjunior:
Unfortunately the auto publish won't work because we have 2fa, I don't think it's be smart to remove it either.

So for now it's still manual as usual

PR #364, opened 2020-12-16
CI: buildah to build images

TriplEight:
Fixes the issue with containerization.

Images manifests are changing with the transition to buildah.

For the reference paritytech/scripts#246


Tbaut:
This project is not maintained @TriplEight

TriplEight:
gotcha

